PR #17, opened 2022-01-18
Github Actions Build

### LINKED ISSUE #12: Create a GitHub Actions pipeline
The pipeline should build the code and run the tests. I recommend separating these two tasks into two different pipeline. 
Also, make sure to configure that pull requests can't be merged if the pipeline fails.

adamv17:
This pull request is intended to add github actions to protect the main branch and make sure that code in the main branch is buildable. closes #12

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -1,0 +1,31 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+# This workflow will build a Java project with Gradle and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Java CI with Gradle
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+      - name: Build with Gradle
+        uses: gradle/gradle-build-action@4137be6a8bf7d7133955359dbd952c0ca73b1021
+        with:
+          arguments: build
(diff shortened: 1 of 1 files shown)


adamv17:
Is there a way to have it be able to run not only on the latest version of windows, so that we don't have problems with this later?

barel13:
I don't think so.
Also, I think you can use wpilib [example](https://docs.wpilib.org/en/stable/docs/software/advanced-gradlerio/robot-code-ci.html)
 for that kind of Action

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,31 @@
...
+# This workflow will build a Java project with Gradle and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-gradle
+
+name: Java CI with Gradle
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
katzuv: Maybe trigger this action for every pull request, not just those opened against `main`.

katzuv:
> Is there a way to have it be able to run not only on the latest version of windows, so that we don't have problems with this later?

Do you mean for Linux too? IMO it's not necessary, just run it on any OS and it should be fine (the code is the same, and if anything the roboRIO is Linux-based). 
(P.S: not sure how it would run on Windows, in Linux you need to grant `gradlew` executing permission (`chmod +x`), not sure if and how it's done on Linux).
I'm seconding @barel13's recommendation, Actions might be painful and it's better to use an existing one. You can also use the Action from 2020 and 2021 repositories.

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,30 @@
+# This is a basic workflow to build robot code.
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the main branch.
```
katzuv: This comment lies.

katzuv (changes requested):

barel13 (approved):

.github/workflows/gradle.yml
adamv17: Done

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,34 @@
+# This is a basic workflow to build robot code.
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow for all push or pull request
```
katzuv:
```suggestion
# Controls when the action will run. Triggers the workflow for all pushes or pull requests.
```

katzuv:
Great! Just make sure you write your own commit messages, GitHub's default "Update ..." isn't meaningful.

katzuv (approved):

PR #24, opened 2022-01-20
Conveyor

adamv17:


src/main/java/frc/robot/subsystems/conveyor/commands/ConveyorDefaultCommand.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    public void end(boolean interrupted) {
+        super.end(interrupted);
+    }
+
+    @Override
+    public boolean isFinished() {
+        return super.isFinished();
+    }
+}
```
katzuv: These declarations are redundant. Do you understand why? 
Also, this command does nothing.

src/main/java/frc/robot/subsystems/conveyor/commands/ConveyorDefaultCommand.java
```diff
@@ -0,0 +1,38 @@
+package frc.robot.subsystems.conveyor.commands;
+
+import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.Ports;
+import frc.robot.subsystems.example.Deez_Nuts;
+
+public class ConveyorDefaultCommand extends CommandBase {
```
katzuv: A command's name needs to describe its action. Also, what if someone wants to use this command without it being the default command?

src/main/java/frc/robot/subsystems/conveyor/commands/ConveyorDefaultCommand.java
```diff
@@ -0,0 +1,38 @@
...
+    private final Deez_Nuts ConveyorDefaultCommand;
```
katzuv: ?

src/main/java/frc/robot/subsystems/conveyor/commands/CargoOut.java
```diff
@@ -0,0 +1,41 @@
...
+        if (conveyor.getColorPosition(true) == conveyor.dr)
+    }
+
+    @Override
+    public void execute() {
+
+
+    }
+
+    // leave me alone
```
katzuv: ?

src/main/java/frc/robot/subsystems/conveyor/commands/CargoOut.java
```diff
@@ -0,0 +1,41 @@
...
+import edu.wpi.first.wpilibj.DriverStation;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.Ports;
+import frc.robot.Constants;
+import frc.robot.subsystems.conveyor.Conveyor;
+
+public class CargoOut extends CommandBase {
+    private final Conveyor conveyor;
+    private final WPI_TalonSRX motor = new WPI_TalonSRX(Ports.Conveyor.AUX);
+//    private frc.robot.subsystems.conveyor.Conveyor.AllianceColor color;
```
katzuv: Don't commit commented-out code.

src/main/java/frc/robot/subsystems/conveyor/commands/CargoOut.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public CargoOut(Conveyor conveyor) {
+        this.conveyor = conveyor;
+        addRequirements(conveyor);
+    }
+
+    @Override
+    public void initialize() {
+        if (conveyor.getColorPosition(true) == conveyor.dr)
```
katzuv: This does nothing.

src/main/java/frc/robot/subsystems/conveyor/commands/CargoOut.java
```diff
@@ -0,0 +1,41 @@
...
+import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX;
+import edu.wpi.first.wpilibj.DriverStation;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.Ports;
+import frc.robot.Constants;
+import frc.robot.subsystems.conveyor.Conveyor;
+
+public class CargoOut extends CommandBase {
+    private final Conveyor conveyor;
+    private final WPI_TalonSRX motor = new WPI_TalonSRX(Ports.Conveyor.AUX);
```
katzuv: See comment for the same line in `CargoIn`.

src/main/java/frc/robot/Constants.java
```diff
@@ -1,10 +1,5 @@
...
-
-    public static class ExampleSubsystem {
-        public static final double POWER = 0.5; // [%]
-    }
-
-}
+package frc.robot;
+
+public class Constants {
+    public static final double POWER = 0.5;
```
katzuv: Why is this in the global constants class?

.idea/modules/robot-2022.iml
```diff
@@ -0,0 +1,13 @@
...
+    <exclude-output />
+    <content url="file://$MODULE_DIR$/../..">
+      <excludeFolder url="file://$MODULE_DIR$/../../.gradle" />
+      <excludeFolder url="file://$MODULE_DIR$/../../build" />
+      <excludeFolder url="file://$MODULE_DIR$/../../out" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
katzuv: Files like this one are autogenerated and shouldn't be committed. Check why Git didn't ignore them.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+import edu.wpi.first.wpilibj.DriverStation;
+import edu.wpi.first.wpilibj.I2C;
+import edu.wpi.first.wpilibj.SPI;
+import edu.wpi.first.wpilibj.util.Color;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Ports;
+
+import java.util.Deque;
+import java.util.LinkedList;
+import com.revrobotics.ColorSensorV3;
```
katzuv: Put this import statement where other stuff from `com.revrobotics`are imported.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+
+public class Conveyor extends SubsystemBase {
+    private static Conveyor INSTANCE = null;
+    private final WPI_TalonSRX motor = new WPI_TalonSRX(Ports.Conveyor.AUX);
+    private DriverStation.Alliance color;
+    private int cargoCount;
```
katzuv: This is redundant, use [`deque.size()`](https://docs.oracle.com/javase/7/docs/api/java/util/Deque.html#size()) instead.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+    private final Deque<String> position = new LinkedList<>();
```
katzuv: Cool! But find a better name.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+    private final ColorSensorV3 colorSensorIntake = new ColorSensorV3(I2C.Port.kMXP);
+    private final ColorSensorV3 colorSensorShooter = new ColorSensorV3(I2C.Port.kOnboard);
```
katzuv: You can use a simple limit switch, beam breaker, etc. instead. No need for a color sensor. You already know the color of the Cargo using the color sensor at the intake.
Also, #40.

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    // leave me alone
+    @Override
+    public boolean isFinished() {
+        if (conveyor.getCargoCount() > 2) {
+            return true;
+        } else {
+            return false;
+        }
```
katzuv:
```suggestion
        return conveyor.getCargoCount() > 2;
```

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,45 @@
...
+    public void initialize() {
+        conveyor.setPower(Constants.POWER);
+    }
+
+    @Override
+    public void execute() {
+
+    }
+
+    // leave me alone
```
katzuv: ?

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    public boolean isFinished() {
+        if (conveyor.getCargoCount() > 2) {
```
katzuv: Magic number.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+    ColorMatch match = new ColorMatch();
```
katzuv: Rename to `colorMatch` or `colorMatcher`, "match" has multiple definitions, especially in FRC.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+    public static Conveyor getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = new Conveyor();
+        }
+        return INSTANCE;
+    }
+
+    public DriverStation.Alliance GetcolorIntake() {
+        Color color = colorSensorIntake.getColor();
+        System.out.println(color.red + " Red " + color.blue + " Blue " + color.green + " Green ");
```
katzuv: This should be called only when debugging, or you can simply delete it.
If it's used only for testing, write a comment to remind yourself to remove it before merging.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+        if (resultColor == Color.kRed) {
+            return DriverStation.Alliance.Red;
+        } else if (resultColor == Color.kBlue) {
+            return DriverStation.Alliance.Blue;
+        } else {
+            return DriverStation.Alliance.Invalid;
+        }
+    }
+
+    public DriverStation.Alliance GetcolorShooter() {
```
katzuv: Rename to `getColorShooter`.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+     * @return the subsystem instance
+     */
+    public static Conveyor getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = new Conveyor();
+        }
+        return INSTANCE;
+    }
+
+    public DriverStation.Alliance GetcolorIntake() {
```
katzuv: Rename to `getColorIntake`.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+        Color resultColor = result.color;
+
+        if (resultColor == Color.kRed) {
+            return DriverStation.Alliance.Red;
+        } else if (resultColor == Color.kBlue) {
+            return DriverStation.Alliance.Blue;
+        } else {
+            return DriverStation.Alliance.Invalid;
+        }
+    }
```
katzuv: `GetcolorIntake` and `GetcolorShooter` are almost identical. You can leave these methods, but extract the logic in them to another, private method.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+
+    public void setPower(double power) {
+        motor.set(power);
+    }
+
+//    public boolean Iscargoalliencecolor() {
+//        return color == Getcolor();
+//    }
+
+    public void setColor(DriverStation.Alliance fmsColor) {
```
katzuv: Remove "fms", the color can come from other sources as well (e.g. Shuffleboard during debugging).

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+        return cargoCount;
+    }
+
+    public void setPower(double power) {
+        motor.set(power);
+    }
+
+//    public boolean Iscargoalliencecolor() {
+//        return color == Getcolor();
+//    }
```
katzuv: Don't commit commented-out code.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+import frc.robot.Ports;
+
+import java.util.Deque;
+import java.util.LinkedList;
+import com.revrobotics.ColorSensorV3;
+
+public class Conveyor extends SubsystemBase {
+    private static Conveyor INSTANCE = null;
+    private final WPI_TalonSRX motor = new WPI_TalonSRX(Ports.Conveyor.AUX);
+    private DriverStation.Alliance color;
```
katzuv: This is never used.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+    public void setColor(DriverStation.Alliance fmsColor) {
+        color = fmsColor;
+    }
+
+    public DriverStation.Alliance getColorPosition(boolean first) {
+        if (first) {
+            return DriverStation.Alliance.valueOf(position.getFirst());
+        }
+        return DriverStation.Alliance.valueOf(position.getLast());
+    }
```
katzuv: Don't repeat yourself:
```suggestion
        String cargoColor;
        if (first) {
            cargoColor = position.getFirst();
        }
        else {
            cargoColor = position.getLast();
        }
        return DriverStation.Alliance.valueOf(cargoColor);
```
Or, better IMO:
```suggestion
        String cargoColor = first ? position.getFirst() : position.getLast();
        return DriverStation.Alliance.valueOf(cargoColor);
```

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+
+//    public boolean Iscargoalliencecolor() {
+//        return color == Getcolor();
+//    }
+
+    public void setColor(DriverStation.Alliance fmsColor) {
+        color = fmsColor;
+    }
+
+    public DriverStation.Alliance getColorPosition(boolean first) {
```
katzuv: Maybe "is intake" instead of "first"?

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+            } else if (motor.getMotorOutputPercent() > 0) {
+                cargoCount++;
+                position.add(enumToString(colorIntake));
+            }
+        }
+    }
+
+    private String enumToString(DriverStation.Alliance alliance) {
+        return alliance.name();
+    }
```
katzuv: Why not just call the code directly where you call `enumToString`?

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+        if (colorIntake != DriverStation.Alliance.Invalid) {
+            if (motor.getMotorOutputPercent() < 0) {
+                cargoCount--;
+                position.removeLast();
+            } else if (motor.getMotorOutputPercent() > 0) {
+                cargoCount++;
+                position.add(enumToString(colorIntake));
+            }
+        }
+    }
```
katzuv: I'm afraid this won't work as expected. If it takes more than 20ms (loop period) for Cargo to trip and un-trip a sensor (which sounds very likely), the Cargo count would be incorrect.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+
+        }
+        if (colorIntake != DriverStation.Alliance.Invalid) {
+            if (motor.getMotorOutputPercent() < 0) {
+                cargoCount--;
+                position.removeLast();
+            } else if (motor.getMotorOutputPercent() > 0) {
+                cargoCount++;
+                position.add(enumToString(colorIntake));
```
katzuv: Use `addLast()` instead to be consistent with `addFirst` (and also so it'd be clearer).

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public void initialize() {
+        conveyor.setPower(Constants.POWER);
+    }
+
+    @Override
+    public void execute() {
+
+    }
```
katzuv: Redundant.
```suggestion
```

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,45 @@
...
+    private final WPI_TalonSRX motor = new WPI_TalonSRX(Ports.Conveyor.AUX);
+
+    public CargoIn(Conveyor conveyor) {
+        this.conveyor = conveyor;
+        addRequirements(conveyor);
+    }
+
+    @Override
+    public void initialize() {
+        conveyor.setPower(Constants.POWER);
```
katzuv: I'm not sure that would work, you might need to repeatedly call it in `execute()`.

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,45 @@
...
+
+    }
+
+    // leave me alone
+    @Override
+    public boolean isFinished() {
+        if (conveyor.getCargoCount() > 2) {
+            return true;
+        } else {
+            return false;
```
katzuv:
```suggestion
        return conveyor.getCargoCount() > 2;
```

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,45 @@
+package frc.robot.subsystems.conveyor.commands;
+
+import com.ctre.phoenix.motorcontrol.can.WPI_TalonSRX;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.Ports;
+import frc.robot.Constants;
+import frc.robot.subsystems.conveyor.Conveyor;
+
+public class CargoIn extends CommandBase {
```
katzuv, line 7: "Intake" is a better name.

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,45 @@
...
+    private final Conveyor conveyor;
+    private final WPI_TalonSRX motor = new WPI_TalonSRX(Ports.Conveyor.AUX);
```
katzuv: I know you're not using this variable (which means it shouldn't be declared), but not only does declaring hardware components outside of a subsystem go against convention, the code will also just crash because you can't declare multiple devices with the same CAN ID.

src/main/java/frc/robot/subsystems/conveyor/commands/CargoOut.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public void initialize() {
+        if (conveyor.getColorPosition(true) == conveyor.dr)
+    }
+
+    @Override
+    public void execute() {
+
+
+    }
```
katzuv: Redundant.
```suggestion
```

src/main/java/frc/robot/subsystems/conveyor/commands/CargoOut.java
```diff
@@ -0,0 +1,41 @@
...
+
+    // leave me alone
+    @Override
+    public boolean isFinished() {
+        return true;
+    }
```
katzuv: Why do you want to end this command immediately?

src/main/java/frc/robot/subsystems/conveyor/commands/CargoOut.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public void end(boolean interrupted) {
+
+    }
```
katzuv:
```suggestion
    
```

src/main/java/frc/robot/subsystems/conveyor/commands/ConveyorDefaultCommand.java
```diff
@@ -0,0 +1,38 @@
...
+    private final WPI_TalonSRX motor = new WPI_TalonSRX(Ports.Conveyor.AUX);
+//    private frc.robot.subsystems.conveyor.Conveyor.AllianceColor color;
```
katzuv: Don't commit commented-out code.

src/main/java/frc/robot/subsystems/conveyor/commands/ConveyorDefaultCommand.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public ConveyorDefaultCommand(Deez_Nuts ConveyorDefaultCommand) {
```
katzuv: Why do you include "default" and "command" here? A command's name should describe what it does, this name doesn't do that.

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    public void end(boolean interrupted) {
+    conveyor.setPower(0);
+    }
+}
+
+
+
```
katzuv, line 34: Redundant newlines.
```suggestion
```

katzuv (changes requested):
Good job!
Please rename the branch's name to lower-case letters.

katzuv:
Additional comments:
1) There are some super long commits in this branch, like bc51b322. Please refrain from doing so, it goes against convention and makes reverting very hard.
2) No need to include "[WIP]", marking the PR as a draft is enough.
3) If a PR is ready for review, it shouldn't be marked as a draft.

src/main/java/frc/robot/Constants.java
```diff
@@ -1,10 +1,9 @@
...
-        public static final double POWER = 0.5; // [%]
-    }
-
-}
+package frc.robot;
+
+public class Constants {
+
+    public static class Conveyor {
+        public static final double POWER = 0.5;
```
barel13: units

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -1,15 +1,21 @@
...
 import edu.wpi.first.wpilibj2.command.button.JoystickButton;
+import frc.robot.subsystems.conveyor.Conveyor;
+import frc.robot.subsystems.conveyor.commands.CargoIn;
+import frc.robot.subsystems.conveyor.commands.CargoOut;
 import frc.robot.valuetuner.ValueTuner;
 import webapp.Webserver;
 
 public class RobotContainer {
   // The robot's subsystems and commands are defined here...
+    public Conveyor conveyor = Conveyor.getInstance();
```
barel13:
```suggestion
    private final Conveyor conveyor = Conveyor.getInstance();
```

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /**
+     *
+     * @param alliance an DriverStation.Alliance enum
+     * @return the name of the enum
+     */
+    private String enumToString(DriverStation.Alliance alliance) {
+        return alliance.name();
+    }
```
barel13: You don't need a function for that, just call `name()`

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,140 @@
...
+            if (motor.getMotorOutputPercent() < 0) {
+                position.removeLast();
+            } else if (motor.getMotorOutputPercent() > 0) {
+                position.add(enumToString(colorIntake));
+            }
+        }
+        lastPassed = hasPassed;
+        lastSeenColor = colorIntake;
+    }
+
```
barel13, line 208: Remove empty line

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,140 @@
...
+    private String enumToString(DriverStation.Alliance alliance) {
+        return alliance.name();
+    }
+
+    /**
+     * removes the string representing the cargo from the list if the cargo is ejected and adds if the cargo is consumed
+     */
+    @Override
+    public void periodic() {
+        var colorIntake = getColor();
```
barel13: You need also to handle the case we outtake through the intake

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
barel13: I don't think so, we would easily confuse with the intake subsystem

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public void execute() {
+
+    }
+
+    // leave me alone
+    @Override
+    public boolean isFinished() {
+        return conveyor.getCargoCount() > Constants.Conveyor.MAX_CARGO_AMOUNT;
```
barel13:
```suggestion
        return conveyor.getCargoCount() >= Constants.Conveyor.MAX_CARGO_AMOUNT;
```

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,40 @@
...
+public class CargoIn extends CommandBase {
+    private final Conveyor conveyor;
+
+    public CargoIn(Conveyor conveyor) {
+        this.conveyor = conveyor;
+        addRequirements(conveyor);
+    }
+
+    @Override
+    public void initialize() {
```
barel13: move to execute

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,40 @@
...
+        conveyor.setPower(Constants.Conveyor.POWER);
```
barel13: I would recommend getting the power as a parameter.

src/main/java/frc/robot/subsystems/conveyor/commands/Convey.java
```diff
@@ -0,0 +1,37 @@
...
+    public Convey(Conveyor conveyor, DoubleSupplier power) {
+        this.conveyor = conveyor;
+        this.power = power;
+        addRequirements(conveyor);
+    }
+
+    @Override
+    public void initialize() {}
+
+
```
barel13: You can remove this method.

barel13 (changes requested):

src/main/java/frc/robot/subsystems/conveyor/commands/Convey.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public boolean isFinished() {
+        return false;
+    }
+
+    @Override
+    public void end(boolean interrupted) {
+    }
```
barel13, line 32: Stop the motor!

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
adamv17: This code does handle this case

adamv17: This code does handle this case

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -1,16 +1,21 @@
...
 import webapp.Webserver;
 
 public class RobotContainer {
   // The robot's subsystems and commands are defined here...
+    private final Conveyor conveyor = Conveyor.getInstance();
     private final XboxController xbox = new XboxController(Ports.Controls.XBOX);
     private final JoystickButton a = new JoystickButton(xbox, XboxController.Button.kA.value);
-
+    private final JoystickButton b = new JoystickButton(xbox, XboxController.Button.kB.value);
+    private final Joystick joystick = new Joystick(0);
```
barel13: add the joystick port to Ports

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -28,7 +33,7 @@ public RobotContainer() {
     }
 
     private void configureDefaultCommands() {
-
+        conveyor.setDefaultCommand(new Convey(conveyor, () -> joystick.getRawAxis(0)));
```
barel13: Please use the constants from XboxController.Axis

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
barel13: also, private final

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,143 @@
...
+    private DriverStation.Alliance lastSeenColor = DriverStation.Alliance.Invalid;
+    private boolean lastPassed = true;
+
+    private Conveyor() {
+        motor.setInverted(TalonFXInvertType.Clockwise);
+        match.addColorMatch(Color.kRed); // red
+        match.addColorMatch(Color.kBlue); // blue
+        match.addColorMatch(Color.kGreen);
+        match.addColorMatch(Color.kYellow);
+        match.addColorMatch(Color.kBlack);
```
barel13: Why do you need so many colors?

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,143 @@
...
+    private final Deque<String> position = new LinkedList<>();
+    private final ColorSensorV3 colorSensorIntake = new ColorSensorV3(I2C.Port.kOnboard);
+    private final DigitalInput beamBreaker = new DigitalInput(Ports.Conveyor.BEAM_BREAKER);
+    private final Solenoid flap = new Solenoid(PneumaticsModuleType.CTREPCM, Ports.Conveyor.SOLENOID);
+    ColorMatch match = new ColorMatch();
+    private DriverStation.Alliance lastSeenColor = DriverStation.Alliance.Invalid;
+    private boolean lastPassed = true;
+
+    private Conveyor() {
+        motor.setInverted(TalonFXInvertType.Clockwise);
```
barel13: Move to ports

barel13 (changes requested):

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
katzuv: > I don't think so, we would easily confuse with the intake subsystem

"cargo in" can be confused with the intake too.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
vichik123: Otherwise it returns blue and red when it shouldn't

src/main/java/frc/robot/Constants.java
```diff
@@ -1,10 +1,14 @@
...
+
+import edu.wpi.first.wpilibj.util.Color;
+
+public class Constants {
+
+    public static class Conveyor {
+        public static final double POWER = 0.5; // [%]
+        public static final int MAX_CARGO_AMOUNT = 2;
+        public static final Color BLUE = new Color(0, 0, 0);
+        public static final Color RED = new Color(0, 0, 0);
```
katzuv: You should already plug here real values (don't forget to tune them on the field at the competition).

src/main/java/frc/robot/Constants.java
```diff
@@ -1,10 +1,14 @@
...
+        public static final Color NONE = new Color(0, 0, 0);
```
katzuv, line 12: Maybe `NONE`'s color should be the color of the surface the color sensor sees when there is no Cargo facing it.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+    private final Solenoid flap = new Solenoid(PneumaticsModuleType.CTREPCM, Ports.Conveyor.SOLENOID);
+    private final ColorMatch match = new ColorMatch();
+    private DriverStation.Alliance lastSeenColor = DriverStation.Alliance.Invalid;
+    private boolean lastPostFlapBeamInput = true;
+
+    private Conveyor() {
+        motor.setInverted(INVERSION);
+        match.addColorMatch(Constants.Conveyor.RED);
+        match.addColorMatch(Constants.Conveyor.BLUE);
+        match.addColorMatch(Constants.Conveyor.NONE);
```
katzuv, line 35: Another, more concise option:
- in the constants file:
```java
 public static final Color[] COLORS =
        {new Color(0, 0, 0),
                new Color(0, 0, 0),
                new Color(0, 0, 0)};
```
- Here:
```java
for (Color color : Constants.Conveyor.COLORS) {
    match.addColorMatch(color);
}
```
That way, if you need to add colors, you only need to modify the `COLORS` array.

---
P.S.: I just noticed you used the colors individually, so this suggestion might not be as helpful as it could. However, it's a nice design pattern you should keep in mind in the future because it can reduce code repetition.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * @return the color sensor value as a DriverStation.Alliance enum.
+     */
+    public DriverStation.Alliance getColor() {
+        Color color = colorSensorIntake.getColor();
+        ColorMatchResult result = match.matchClosestColor(color);
+        Color resultColor = result.color;
+        SmartDashboard.putBoolean("match", resultColor == Color.kRed);
```
katzuv, line 59: This is not a good name for the field. It's not clear whether `color is blue` means `match == true` or the other way around. You should rename it to something more specific.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+            return DriverStation.Alliance.Red;
+        } else if (resultColor == Constants.Conveyor.BLUE) {
+            return DriverStation.Alliance.Blue;
+        } else {
+            return DriverStation.Alliance.Invalid;
+        }
+    }
+
+    /**
+     * @return the amount of balls inside the cargo.
```
katzuv, line 71:
```suggestion
     * @return the amount of Cargo inside the conveyor.

```

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+     * closes the "flap" - solenoid
+     */
+    public void closeFlap() {
+        flap.set(true);
+    }
+
+    /**
+     * toggles the "flap" - solenoid
+     */
+    public void toggle() {
```
katzuv: Be specific and consistent – rename to `toggleFlap`.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * closes the "flap" - solenoid
+     */
+    public void closeFlap() {
+        flap.set(true);
+    }
+
+    /**
+     * toggles the "flap" - solenoid
```
katzuv, line 101: "solenoid" shouldn't appear here, it's a hardware implementation. Remember, documentation should describe _what_ your class or function does, not _how_ they do it unless it's important for the user to know that.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+     */
+    public void toggle() {
+        flap.toggle();
+    }
+
+    private void updateActualBallPositions() {
+        var colorIntake = getColor();
+        boolean currentPostFlapBeamInput = postFlapBeam.get();
+        SmartDashboard.putString("alliance", colorIntake.name());
```
katzuv, line 110: "alliance" sounds like the color of the alliance _Stardust_ plays in. That's not the case here though.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+        Case #5  queue = [Alliance, Opponent]
+        Case #6  queue = [None, Opponent]
+        Case #7  queue = [Opponent, None]
+        Case #8  queue = [Opponent, Opponent]
+        Case #9  queue = [Opponent, Alliance]
+         */
+        var last = position.getLast();
+        var first = position.getFirst();
+        var invalid = DriverStation.Alliance.Invalid.name();
+        var alliance = DriverStation.getAlliance().name();
```
katzuv: 1. invalid what? alliance what?
2. `invalid` and `alliance` can be declared at the subsystem's constructor, no need to create them repeatedly.
3. Can't `invalid` be replaced with `Constants.Conveyor.NONE`?

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+                return Queue.AllianceAlliance;
+            }
+            return Queue.AllianceOpponent;
+        }
+        if (first.equals(invalid)) {
+            return Queue.OpponentNone;
+        } else if (first.equals(alliance)) {
+            return Queue.OpponentAlliance;
+        }
+        return Queue.OpponentOpponent;
```
katzuv: The diff doesn't render very nicely but I deleted all the `else if`s and replaced them with `simplify if`s. Do you understand why?
~~also please check if I'm correct it's 02:48 AM~~
```suggestion
        if (last.equals(invalid)) {
            if (first.equals(invalid)) {
                return Queue.NoneNone;
            } if (first.equals(alliance)) {
                return Queue.NoneAlliance;
            }
            return Queue.NoneOpponent;
        }
        if (last.equals(alliance)) {
            if (first.equals(invalid)) {
                return Queue.AllianceNone;
            } if (first.equals(alliance)) {
                return Queue.AllianceAlliance;
            }
            return Queue.AllianceOpponent;
        }
        if (first.equals(invalid)) {
            return Queue.OpponentNone;
        } if (first.equals(alliance)) {
            return Queue.OpponentAlliance;
        }
        return Queue.OpponentOpponent;
```

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+        Case #3  queue = [Alliance, None]
+        Case #4  queue = [Alliance, Alliance]
+        Case #5  queue = [Alliance, Opponent]
+        Case #6  queue = [None, Opponent]
+        Case #7  queue = [Opponent, None]
+        Case #8  queue = [Opponent, Opponent]
+        Case #9  queue = [Opponent, Alliance]
+         */
+        var last = position.getLast();
+        var first = position.getFirst();
```
katzuv: 1. Last what?
2. First what?
3. On _Stardust_, where is "first" and where is "last"?
4. why is `last` declared` before `first`?

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -0,0 +1,198 @@
...
+    public enum Queue {
+        NoneNone(new int[]{0, 0}),
+        NoneAlliance(new int[]{0, 1}),
+        AllianceNone(new int[]{1, 0}),
+        AllianceAlliance(new int[]{1, 1}),
+        AllianceOpponent(new int[]{1, 2}),
+        NoneOpponent(new int[]{0, 2}),
+        OpponentNone(new int[]{2, 0}),
+        OpponentOpponent(new int[]{2, 2}),
+        OpponentAlliance(new int[]{2, 1});
```
katzuv, line 224: Very cool!

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
```diff
@@ -0,0 +1,34 @@
...
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.Constants;
+import frc.robot.subsystems.conveyor.Conveyor;
+
+public class CargoIn extends CommandBase {
+    private final Conveyor conveyor;
+    private final double power;
+
+    public CargoIn(Conveyor conveyor, double power) {
```
katzuv, line 11: Maybe rename to `conveyCargoIn`, or better, `ConveyCargoUp`, because commands' names should start with a verb.

src/main/java/frc/robot/subsystems/conveyor/commands/Convey.java
```diff
@@ -0,0 +1,34 @@
+package frc.robot.subsystems.conveyor.commands;
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.conveyor.Conveyor;
+
+import java.util.function.DoubleSupplier;
+
+public class Convey extends CommandBase {
```
katzuv, line 8: 1. This name is confusing, what's the difference between "Cargo in" and "convey"?
2. These two commands are very similar. Perhaps you should combine them into a single command?

katzuv (changes requested):
Magnificent!

katzuv:
How late is it? It's The Cheesy Poofs time 😢.

src/main/java/frc/robot/Constants.java
barel13: That's exactly the purpose

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
barel13: It's just for now for testing purposes.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
barel13: Again, it's only for testing purposes, we will delete it later.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
barel13: 1/2. the position in the queue.
3. first is the first one to come into the conveyor.
4. because it doesn't matter.

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
barel13: 1. invalid alliance (in this context it means empty).
2. It doesn't really matter, it's just for making the lines shorter.
3. Nope, we compare the alliance's color (not the same type)

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
barel13: He knows he doesn't need the `else if`, he said the code look better to him with the `else if`

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
barel13: Just add comment, something like this:
```
0 - None.
1 - Alliance.
2 - Opponent.
```

src/main/java/frc/robot/subsystems/conveyor/commands/CargoIn.java
barel13: OK, I still don't think naming it `Intake` is a better name.

barel13 (commented):
You can use this function to make sure only the 2 latest results will remain:
```java
        var queue = new ArrayBlockingQueue<...>(2);
        if (queue.remainingCapacity() == 0) {
            queue.poll();
        }
        queue.offer(value);
```

PR #27, opened 2022-01-20
Hood

adamv17:


src/main/java/frc/robot/Constants.java
```diff
@@ -2,9 +2,14 @@
...
-
-    public static class ExampleSubsystem {
-        public static final double POWER = 0.5; // [%]
+    public static final double LOOP_PERIOD = 0.02; // loop period. [s]
+    public static final double g = 9.80665; // Gravity acceleration constant. [m/s^2]
+    public static final double UPPER_TARGET_HEIGHT = 2.64; // Height of upper target. [m]
+    public static final double PEBZNER_HEIGHT = 4.8; // Height of pebzner auditorium. [m]
+    public static final int TALON_TIMEOUT = 10; // Waiting period for configurations [ms].
+
+    public static class AngleChanger {
```
adamv17: No, change to Hood

src/main/java/frc/robot/Ports.java
```diff
@@ -1,13 +1,11 @@
...
 
 public final class Ports {
-    public static class ExampleSubsystem {
-        public static final int MAIN = 0;
-        public static final int AUX = 0;
-        public static final boolean IS_MAIN_INVERTED = false;
-        public static final boolean IS_AUX_INVERTED = false;
-        public static final boolean IS_MAIN_SENSOR_INVERTED = false;
-        public static final boolean IS_AUX_SENSOR_INVERTED = false;
+    public static class Shooter {
```
adamv17: Create a Hood class instead

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,27 @@
+package frc.robot.subsystems.hood;
+
+import edu.wpi.first.wpilibj.PneumaticsModuleType;
+import edu.wpi.first.wpilibj.Solenoid;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+
+import static frc.robot.Ports.Shooter.SOLENOID;
+
+public class Hood extends SubsystemBase {
+    private static final Hood INSTANCE = new Hood();
```
adamv17: Use lazy initialization

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public Hood() {
+    }
+
+    public static Hood getINSTANCE() {
+        return INSTANCE;
+    }
+
+    public void changeAngle() {
+        angleChanger.set(!angleChanger.get());
```
adamv17: Use toggle

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
```
adamv17, line 59: Add functions open and close

src/main/java/frc/robot/subsystems/hood/commands/HoodDefaultCommand.java
```diff
@@ -0,0 +1,41 @@
...
+        addRequirements(hood);
+    }
+
+    @Override
+    public void initialize() {
+    }
+
+    @Override
+    public void execute() {
+        input = modeSupplier.get();
```
adamv17: Use the open close functions instead

src/main/java/frc/robot/subsystems/hood/commands/HoodDefaultCommand.java
```diff
@@ -0,0 +1,41 @@
+package frc.robot.subsystems.hood.commands;
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.subsystems.hood.Hood;
+
+import java.util.function.Supplier;
+
+public class HoodDefaultCommand extends CommandBase {
```
adamv17: Don't make it a default command, just an InstantCommand

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,73 @@
...
+import static frc.robot.Ports.Hood.SOLENOID;
+
+public class Hood extends SubsystemBase {
+    private static Hood INSTANCE;
+    private final Solenoid angleChanger = new Solenoid(PneumaticsModuleType.CTREPCM, SOLENOID);
+
+    public Hood() {
+    }
+
```
barel13: private

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Opens the solenoid.
+     * This function simply toggles the mode of the
+     * solenoid only if it's currently closed.
+     */
+    public void open() {
+        if (angleChanger.get())
+            angleChanger.toggle();
+
```
barel13: Just use angleChanger.set(true);

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Closes the solenoid.
+     * This function simply toggles the mode of the
+     * solenoid only if it's currently open.
+     */
+    public void close() {
+        if (!angleChanger.get()) {
+            angleChanger.toggle();
+        }
+    }
```
barel13: Same as above

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,73 @@
...
+    public void setSolenoid(Mode mode){
+        angleChanger.set(mode.value);
+    }
+
+    /**
+     * Returns whether the solenoid is active.
+     *
+     * @return a boolean representing the mode of the solenoid.
+     */
+    public boolean getActive() {
```
barel13: So maybe rename it to `isActive()`?

src/main/java/frc/robot/Constants.java
```diff
@@ -2,9 +2,14 @@
...
 public final class Constants {
-
-    public static class ExampleSubsystem {
-        public static final double POWER = 0.5; // [%]
+    public static final double LOOP_PERIOD = 0.02; // loop period. [s]
+    public static final double g = 9.80665; // Gravity acceleration constant. [m/s^2]
+    public static final double UPPER_TARGET_HEIGHT = 2.64; // Height of upper target. [m]
+    public static final double PEBZNER_HEIGHT = 4.8; // Height of pebzner auditorium. [m]
+    public static final int TALON_TIMEOUT = 10; // Waiting period for configurations [ms].
+
```
barel13: You don't need these constants for the Hood, delete them

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -2,17 +2,20 @@
...
 
 public class RobotContainer {
-  // The robot's subsystems and commands are defined here...
-    public ExampleSubsystem exampleSubsystem = ExampleSubsystem.getInstance();
+    // The robot's subsystems and commands are defined here...
     private final XboxController xbox = new XboxController(Ports.Controls.XBOX);
     private final JoystickButton a = new JoystickButton(xbox, XboxController.Button.kA.value);
-
+    private final JoystickButton b = new JoystickButton(xbox, XboxController.Button.kB.value);
+    private final Trigger rt = new Trigger(() -> xbox.getRightTriggerAxis() > 0);
```
barel13: The number should be higher than 0, and use a constant

barel13 (changes requested):

src/main/java/frc/robot/Constants.java
```diff
@@ -3,8 +3,8 @@
 
 public final class Constants {
 
-    public static class ExampleSubsystem {
-        public static final double POWER = 0.5; // [%]
+    public static class Hood {
+        public static final double CLOSED_ANGLE = 54; // Angle for longer distance. [deg]
+        public static final double OPEN_ANGLE = 69; // Angle for shorter distance. [deg]
```
katzuv: These aren't used anywhere.
```suggestion
```

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -30,21 +32,21 @@ public RobotContainer() {
...
-    *
-    * @return the command to run in autonomous
-    */
+     * Use this to pass the autonomous command to the main {@link Robot} class.
+     *
+     * @return the command to run in autonomous
+     */
     public Command getAutonomousCommand() {
-      return null;
+        return null;
```
katzuv: This change is made in too many branches. Instead, create a small PR for formatting fixes. Then, make sure you format your code before committing it. [This can be done easily with IntelliJ IDEA's excellent Git integration](https://www.jetbrains.com/help/idea/commit-and-push-changes.html#commit) and will prevent redundant diffs.

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Closes the solenoid.
+     * This function simply toggles the mode of the
+     * solenoid only if it's currently open.
+     */
+    public void close() {
+        angleChanger.set(true);
+    }
```
katzuv, line 54: But this is an angle changer, it doesn't open or close anything.

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public void setSolenoid(Mode mode){
+        angleChanger.set(mode.value);
+    }
+
+    /**
+     * Returns whether the solenoid is active.
```
katzuv: What does "active" mean here?

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @return a boolean representing the mode of the solenoid.
```
katzuv:
```suggestion
     * @return whether the solenoid is active.
```
However, "solenoid" shouldn't be here, don't expose hardware components out.

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -0,0 +1,68 @@
...
+public class Hood extends SubsystemBase {
+    private static Hood INSTANCE;
+    private final Solenoid angleChanger = new Solenoid(PneumaticsModuleType.CTREPCM, SOLENOID);
+
+    private Hood() {
+    }
+
+    public enum Mode {
+        ShortDistance(true),
+        LongDistance(false);
```
katzuv, line 18: I don't think true and false fit here. Why not angles?

katzuv (changes requested):

src/main/java/frc/robot/Constants.java
vichik123: They may be used for trajectory calculations. It's better to leave them there so that we don't forget.

src/main/java/frc/robot/RobotContainer.java
vichik123: I'm having trouble deleting the extra spaces.

src/main/java/frc/robot/subsystems/hood/Hood.java
vichik123: The solenoid opens and closes according to the desired angle

src/main/java/frc/robot/subsystems/hood/Hood.java
vichik123: It represents the 2 modes of the solenoid. true - open, false - closed.

src/main/java/frc/robot/RobotContainer.java
katzuv: First, if you're having trouble – why did you resolve the conversation?
Second, what is the problem?

src/main/java/frc/robot/Constants.java
katzuv: > It's better to leave them there so that we don't forget.

That's bad practice. Leaving unused code is bad and will clutter your codebase.
This code is committed. You can safely delete it and revert the deletion if needed in the future.
And even if they'll be used, they'll probably be in the shooter's constants, not here.

src/main/java/frc/robot/subsystems/hood/Hood.java
katzuv: A boolean doesn't fit here IMO. It's not like in the intake where "retracted" is the opposite of "un-retracted" (and even there I would use an enum). Here, 54° is not the opposite of 69°. They are just two different options. What would you do if you had three options?

src/main/java/frc/robot/RobotContainer.java
vichik123: I figured it out. Turns out I just needed to highlight the spaces 😅.

src/main/java/frc/robot/subsystems/hood/Hood.java
vichik123: There can't be 3 options because the only part used in the subsystem is a single solenoid.

barel13 (approved):
Ctrl+Alt+L

Se-gol (approved):

PR #48, opened 2022-01-28
Climber feature

GaiaZano05:


barel13 (approved):

barel13 (approved):
WOW

src/main/java/frc/robot/Constants.java
```diff
@@ -49,6 +49,9 @@
 
         public static final double STOP_CLIMBER = 149.5; // [s]
 
+        public static final double SIMULATION_LOOP_PERIOD = 0.02; // [s]
```
katzuv, line 52: Isn't there already a "loop period" constant?

src/main/java/frc/robot/Constants.java
```diff
@@ -49,6 +49,9 @@
...
+
```
katzuv, line 53: One line is enough.
```suggestion
```

katzuv (changes requested):
Rename the branch's name to `feature/climber` ([here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch)'s how to do that).

src/main/java/frc/robot/Constants.java
barel13: not for the simulations

barel13:
I think it shouldn't be in Climber's constants, its related also to other subsystems

PR #54, opened 2022-01-28
Update gitignore

### LINKED ISSUE #53: Update gitignore
Waits until the subsystems are merged integrated

barel13:
Comment


katzuv:
1) What's the reason to wait until all subsystems are merged?
2) Why are there changes to files other than `.gitignore`?

barel13:
1. Because I think other branches will override if they merge after this branch.
2. Because we don't need to keep the files with the name simgui (part of wpilib simulation files, overriden after altering the gui)

katzuv:
1) What do you mean by "override"?
2) I know what they are, but the issue's title is "Update gitignore", so they should be deleted in another pull request.
3) If you want a review, you should mark the PR as "Ready to review". FTFY.

barel13:
1. Because they have a "new" addition to the file, git might choose their version of the file we just removed.
 2. I don't think the same, because I updated the gitignore we also need to apply the changes, and the changes are directly related to this change.
 3. I know it's a draft. As I said before, I afraid other branches will override the changes, so I don't want anyone to merge this branch until I'm sure there won't be a problem

katzuv:
1) It shouldn't work that way. You may have conflicts, and in the worst case, you can `git revert` back to happiness. If you have concerns, you can make all merges local (then revert) and then you'll see if you have any unsolvable problems.
2) OK, you're right. Didn't think about it.

katzuv (approved):

adamv17 (approved):

PR #49, opened 2022-01-28
Shooter

vichik123:
Had to rename the branch and the previous PR was deleted.

PR #59, opened 2022-02-01
Pipeline of Cargo

vichik123:
Leave a comment


src/main/java/frc/robot/subsystems/shooter/Shooter.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /**
+     * Calculates the velocity setpoint according to the distance from the target.
+     * Once the data from the shooter is acquired this function will be changed.
+     *
+     * @param distance is the distance from the target. [m]
+     * @return 15. [rpm]
+     */
+    public static double getSetpointVelocity(double distance) {
```
katzuv: Continuing the conversation from [here](https://github.com/Galaxia5987/robot-2022/pull/49#discussion_r795997098):
> The function will be based off of the values that we measure with our testing. This objectively means that the interpolation belongs to the Shooter subsystem itself, and not the command. (@littlebro123)

I'm not sure I understand. Why does it matter where do the values come from? Would it be different if they would come from, say,  theoretical calculations?

adamv17: I agree that it makes more sense to have this method in the command, shooter can be seen as a module that receives a velocity and gets to it to shoot. The command is what decides that velocity and the command receives the distance.

adamv17 (approved):

barel13 (approved):

PR #61, opened 2022-02-01
Feature/climber

### LINKED ISSUE #52: Use the absolute encoder on the climber
We have absolute encoder on the climber

GaiaZano05:
The code of the subsystem Climber.

motors: TalonFX * 2

Commands:
- AdjustAngle
- ClimberStop
- JoystickClimb

closes #52


src/main/java/frc/robot/Constants.java
```diff
@@ -49,25 +53,33 @@
...
-                    new Translation2d(0.038, 0.171), new Rotation2d()); // Position of the vision relative to the robot.
-
-            public static final SimVisionTarget SIM_TARGET_HUB = new SimVisionTarget( // Hub target for vision simulation.
-                    HUB_POSE, TARGET_HEIGHT_FROM_GROUND, TARGET_WIDTH, TARGET_HEIGHT);
-        }
+        public static final double SIMULATION_LOOP_PERIOD = 0.02; // [s]
+
+        public static final int ZERO_POSITION = 0;
+
+        public static final int ZERO_POSITION_TOLERANCE = 0;
```
barel13: units

src/main/java/frc/robot/subsystems/climber/Climber.java
```diff
@@ -149,22 +156,30 @@ public void setVelocity(double velocity) {
             double volts = controller.calculate(getVelocity(), velocity);
             mainMotor.setVoltage(volts + feedforward.calculate(0, 0));
         } else {
-            mainMotor.set(ControlMode.Velocity, unitModel.toTicks100ms(velocity));
+            mainMotor.set(ControlMode.Velocity, unitModelPosition.toTicks100ms(velocity));
         }
     }
 
+
+    public void setAngleZero() {
```
barel13, line 170: documentation

src/main/java/frc/robot/subsystems/climber/Climber.java
```diff
@@ -149,22 +156,30 @@ public void setVelocity(double velocity) {
...
+        setPosition(angle);
+    }
+
+
     /**
      * @return get motors position. [rad]
      */
     public double getPosition() {
-        return unitModel.toUnits(mainMotor.getSelectedSensorPosition());
+        return unitModelPosition.toUnits(mainMotor.getSelectedSensorPosition(1));
```
barel13:
```suggestion
        return unitModelPosition.toUnits(mainMotor.getSelectedSensorPosition(0));
```

src/main/java/frc/robot/subsystems/climber/commands/AdjustAngle.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public void execute() {
+        climber.setAngleZero();
+    }
+
+    @Override
+    public boolean isFinished() {
+        return Math.abs(climber.getPosition() - Constants.Climber.ZERO_POSITION)  >= Constants.Climber.ZERO_POSITION_TOLERANCE;
+    }
```
barel13, line 29: stop the motor at the end

barel13 (changes requested):

src/main/java/frc/robot/Constants.java
```diff
@@ -35,6 +37,8 @@
         public static final double GEAR_RATIO = 292.1;
 
         public static final double TICKS_PER_RAD = 2048 * GEAR_RATIO / (2 * Math.PI);
+        public static final double TICKS_PER_MOTOR =  2048;
```
katzuv: per _motor_?

src/main/java/frc/robot/subsystems/climber/Climber.java
```diff
@@ -37,13 +38,16 @@
     private final WPI_TalonFX mainMotor = new WPI_TalonFX(Ports.Climber.MAIN);
     private final WPI_TalonFX auxMotor = new WPI_TalonFX(Ports.Climber.AUX);
```
katzuv, line 39: This can be a local variable inside the constructor, there's no need to store it as a class member.

src/main/java/frc/robot/subsystems/climber/Climber.java
```diff
@@ -81,6 +85,8 @@ private Climber() {
             armTower.setColor(new Color8Bit(Color.kBlue));
         }
 
+        mainMotor.configSelectedFeedbackSensor(FeedbackDevice.RemoteSensor0, 1, Constants.TALON_TIMEOUT);
```
katzuv, line 87: Why not use the primary loop?

src/main/java/frc/robot/subsystems/climber/Climber.java
```diff
@@ -96,9 +102,10 @@ private Climber() {
          */
         mainMotor.configMotionCruiseVelocity(Constants.Climber.CRUISE_VELOCITY);
         mainMotor.configMotionAcceleration(Constants.Climber.MAXIMAL_ACCELERATION);
-        mainMotor.config_kP(0, Constants.Climber.KP);
-        mainMotor.config_kI(0, Constants.Climber.KI);
-        mainMotor.config_kD(0, Constants.Climber.KD);
+        mainMotor.config_kP(0, Constants.Climber.KP, Constants.TALON_TIMEOUT);
+        mainMotor.config_kI(0, Constants.Climber.KI, Constants.TALON_TIMEOUT);
+        mainMotor.config_kD(0, Constants.Climber.KD, Constants.TALON_TIMEOUT);
```
katzuv: If you're not reading the error code returned from the config functions, the timeout has no use.

src/main/java/frc/robot/subsystems/climber/Climber.java
```diff
@@ -37,13 +38,16 @@
     private final WPI_TalonFX mainMotor = new WPI_TalonFX(Ports.Climber.MAIN);
     private final WPI_TalonFX auxMotor = new WPI_TalonFX(Ports.Climber.AUX);
     private final Solenoid stopper = new Solenoid(PneumaticsModuleType.CTREPCM, Ports.Climber.STOPPER);
-    private final UnitModel unitModel = new UnitModel(Constants.Climber.TICKS_PER_RAD);
+    private final UnitModel unitModelGearbox = new UnitModel(Constants.Climber.TICKS_PER_MOTOR);
+    private final UnitModel unitModelPosition = new UnitModel(Constants.Climber.TICKS_PER_RAD);
```
katzuv: What's the difference between the two?

src/main/java/frc/robot/subsystems/climber/Climber.java
katzuv: Especially document what zero means (e.g. parallel or perpendicular to the field).

src/main/java/frc/robot/Constants.java
katzuv: You'll probably never be _exactly_ on point. Tolerance should be greater than zero (although not by a lot).

src/main/java/frc/robot/subsystems/climber/Climber.java
```diff
@@ -149,22 +156,30 @@ public void setVelocity(double velocity) {
...
         } else {
-            mainMotor.set(ControlMode.Velocity, unitModel.toTicks100ms(velocity));
+            mainMotor.set(ControlMode.Velocity, unitModelPosition.toTicks100ms(velocity));
         }
     }
 
+
+    public void setAngleZero() {
+        double angle = unitModelGearbox.toUnits(mainMotor.getSelectedSensorPosition(1)
+                - Constants.Climber.ZERO_POSITION);
```
katzuv: `ZERO_POSITION = 0` so this does nothing. Should it just be `setPosition(onstants.Climber.ZERO_POSITION)`?

src/main/java/frc/robot/subsystems/climber/commands/AdjustAngle.java
katzuv: Another, more elegant option IMO, is to call [`configAllowableClosedloopError()`](https://store.ctr-electronics.com/content/api/java/html/classcom_1_1ctre_1_1phoenix_1_1motorcontrol_1_1can_1_1_base_motor_controller.html#a940631668a8428481eff932f350046d5). That way, the motor will stop automatically when the error reaches the configured allowable value.
Note: although I'm 99% sure this will set the motor to the neutral mode (=brake) ([source](https://docs.ctre-phoenix.com/en/stable/ch16_ClosedLoop.html?highlight=Allowable#closed-loop-configs-per-slot-four-slots-available)), you should verify it.

src/main/java/frc/robot/Constants.java
```diff
@@ -49,25 +51,33 @@
...
-                    new Translation2d(0.038, 0.171), new Rotation2d()); // Position of the vision relative to the robot.
-
-            public static final SimVisionTarget SIM_TARGET_HUB = new SimVisionTarget( // Hub target for vision simulation.
-                    HUB_POSE, TARGET_HEIGHT_FROM_GROUND, TARGET_WIDTH, TARGET_HEIGHT);
-        }
+        public static final double SIMULATION_LOOP_PERIOD = 0.02; // [s]
+
+        public static final int ZERO_POSITION = 0; //[radians]
+
+        public static final int ZERO_POSITION_TOLERANCE = 0; //[radians]
```
adamv17: Change tolerance to reasonable constant

src/main/java/frc/robot/subsystems/climber/commands/AdjustAngle.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public void execute() {
+        climber.setAngleZero();
+    }
+
+    @Override
+    public boolean isFinished() {
+        return Math.abs(climber.getPosition() - Constants.Climber.ZERO_POSITION)  >= Constants.Climber.ZERO_POSITION_TOLERANCE;
```
katzuv: Maybe I don't understand something, but why are you subtracting `ZERO_POSITION` from the current position?

src/main/java/frc/robot/subsystems/climber/Climber.java
katzuv, line 87: The CTRE docs say ([source](https://docs.ctre-phoenix.com/en/stable/ch14a_BringUpRemoteSensors.html#bring-up-remote-sensors)):
> The remote sensor filter feature of the Talon SRX/Victor SPX allows for the processing of sensor values provided from other remote CAN bus devices. In other words, a Talon SRX or Victor SPX can execute closed-loop modes with sensor values **sourced by other Talons, CANifiers, or Pigeon IMUs on the same CAN-bus**. (emphasis mine)

AFAIK you plan to use REV Robotics' Through Bore Encoder which doesn't have a CAN-bus connector. Assuming you don't plan to use a CANifier, how are you going to use the encoder as a remote sensor?

adamv17 (approved):

src/main/java/frc/robot/Constants.java
```diff
@@ -32,13 +32,11 @@
         public static final double MAXIMAL_ACCELERATION = 0; // [ticks/100ms*sec]
         public static final boolean VOLTAGE_COMPENSATION = true;
 
-        public static final double MAX_VELOCITY = Math.PI * 2 / 3; //[rad/s]
+        public static final double MAX_VELOCITY = Math.PI * 2 / 3; //[rad// /s]
```
katzuv: Why? `rad/s` is good 🙂.

katzuv (changes requested):

katzuv:
A few notes about the commands which somehow don't appear here:
- Maybe rename `ClimberStop` to `StopClimber`, sounds more natural to me.

https://github.com/Galaxia5987/robot-2022/blob/4fcd2a5891aa829a92e40b785f8393ab222d8bae/src/main/java/frc/robot/subsystems/climber/commands/JoystickClimb.java#L35 I don't think `Timer.getFPGATimestamp()` will give you the desired result. The documentation says this function returns "the time since the FPGA started". The roboRIO boots up before the match starts so you can't know what the FPGA timestamp will be when the match ends. Instead, I recommend using [`Timer.getMatchTime()`](https://first.wpi.edu/wpilib/allwpilib/docs/release/java/src-html/edu/wpi/first/wpilibj/Timer.html#line.34). Read the docs carefully though, it doesn't necessarily return the value you expect.
https://github.com/Galaxia5987/robot-2022/blob/4fcd2a5891aa829a92e40b785f8393ab222d8bae/src/main/java/frc/robot/subsystems/climber/commands/ClimberStop.java#L22 Why is this condition necessary? Nothing bad will happen if the command will run in simulation.
https://github.com/Galaxia5987/robot-2022/blob/4fcd2a5891aa829a92e40b785f8393ab222d8bae/src/main/java/frc/robot/subsystems/climber/commands/ClimberStop.java#L23-L25 Why do you toggle the stopper in that case?

src/main/java/frc/robot/Constants.java
```diff
@@ -49,25 +51,33 @@
 
         public static final double STOP_CLIMBER = 149.5; // [s]
```
katzuv: Add "timestamp" or something similar to this constant's name. It's not clear whether this is a time diff or a target angle.

src/main/java/frc/robot/subsystems/climber/commands/AdjustAngle.java
```diff
@@ -0,0 +1,30 @@
+package frc.robot.subsystems.climber.commands;
+
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.Constants;
+import frc.robot.subsystems.climber.Climber;
+
+public class AdjustAngle extends CommandBase {
```
katzuv, line 7: What is the purpose of this command?

src/main/java/frc/robot/subsystems/climber/Climber.java
barel13: We don't write motors this way as it's more easier to read that way. Furthermore, for other people who read our code, it will be a lot more easier to know how much motors we use this way.

src/main/java/frc/robot/subsystems/climber/Climber.java
barel13: Because we use the internal sensor for that

src/main/java/frc/robot/subsystems/climber/Climber.java
barel13: It's just a convention to add the timeout

src/main/java/frc/robot/subsystems/climber/Climber.java
barel13: No, the zero position changes according to the position of the absolute encoder, which we can't know right now without the robot...

barel13: This function gets the absolute angle of the motor (relative to the horizon), and then we move the motor (using the internal encoder)

src/main/java/frc/robot/subsystems/climber/commands/AdjustAngle.java
barel13: `ZERO_POSITION` = the position where the angle is zero

src/main/java/frc/robot/subsystems/climber/Climber.java
adamv17: You might be right we will check with the captain how the remote sensor is connected

src/main/java/frc/robot/subsystems/climber/commands/AdjustAngle.java
barel13: IMO we should rename this command to something like `StabilizeClimber`. The purpose of this command is to make the climber horizontal (so we will be in the frame perimeter at the beginning of the match)

src/main/java/frc/robot/Constants.java
adamv17: I believe it was mistyped and not on purpose

src/main/java/frc/robot/Constants.java
```diff
@@ -5,9 +5,11 @@
 import edu.wpi.first.math.geometry.Rotation2d;
 import edu.wpi.first.math.geometry.Transform2d;
 import edu.wpi.first.math.geometry.Translation2d;
+import frc.robot.subsystems.UnitModel;
```
barel13: You can remove this import

src/main/java/frc/robot/Constants.java
```diff
@@ -49,25 +51,33 @@
...
-
-            public static final Pose2d HUB_POSE = new Pose2d( // Position of the hub relative to the field.
-                    new Translation2d(FIELD_LENGTH / 2, FIELD_WIDTH / 2), new Rotation2d());
-            public static final Transform2d CAMERA_TO_ROBOT = new Transform2d(
-                    new Translation2d(0.038, 0.171), new Rotation2d()); // Position of the vision relative to the robot.
-
-            public static final SimVisionTarget SIM_TARGET_HUB = new SimVisionTarget( // Hub target for vision simulation.
-                    HUB_POSE, TARGET_HEIGHT_FROM_GROUND, TARGET_WIDTH, TARGET_HEIGHT);
-        }
+        public static final double SIMULATION_LOOP_PERIOD = 0.02; // [s]
```
barel13: You can move it to `Constants` (not `Constants.Climber`)

src/main/java/frc/robot/Robot.java
```diff
@@ -5,9 +5,8 @@
 package frc.robot;
 
 import com.kauailabs.navx.frc.AHRS;
-import edu.wpi.first.wpilibj.PowerDistribution;
-import edu.wpi.first.wpilibj.SPI;
-import edu.wpi.first.wpilibj.TimedRobot;
+import edu.wpi.first.wpilibj.*;
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
```
barel13, line 9: Ctrl+Alt+L

barel13 (changes requested):

barel13 (approved):

Issue #37, opened 2022-01-21
Adam is alive


barel13:
die boi die

katzuv:
Please open a pull request to fix the issue.

vichik123:
This issue doesn't really involve any code.

katzuv:
> This issue doesn't really involve any code.

What about path following, with Adam being on one of the waypoints?

mandelete:
hmm very intewesting

adamv17:
This is fine

mandelete:
ah yes
die

vichik123:
> > This issue doesn't really involve any code.
> 
> What about path following, with Adam being on one of the waypoints?

The path weaving usually doesn't allow for high speeds, so unaliving Adam by having him as one of the waypoints isn't very effective.

vichik123:
Lady and gentlemen the issue is resolved.

PR #66, opened 2022-02-05
Feature/pneumatics base

adamv17:
Add the pneumatics base and utilize it to check whether the solenoids are connected.

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -5,10 +5,7 @@
 package frc.robot;
 
 import com.kauailabs.navx.frc.AHRS;
-import edu.wpi.first.wpilibj.DriverStation;
-import edu.wpi.first.wpilibj.PowerDistribution;
-import edu.wpi.first.wpilibj.SPI;
-import edu.wpi.first.wpilibj.TimedRobot;
+import edu.wpi.first.wpilibj.*;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.CommandScheduler;
 /**
@@ -23,6 +20,7 @@ public class Robot extends TimedRobot {
     public PowerDistribution pdp = new PowerDistribution();
     private RobotContainer m_robotContainer;
     private Command m_autonomousCommand;
+    public static final PneumaticsBase pneumaticsBase = PneumaticsBase.getForType(0, PneumaticsModuleType.CTREPCM);
 
     /**
      * This function is run when the robot is first started up and should be used for any
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -5,7 +5,11 @@
 package frc.robot;
 
 import com.kauailabs.navx.frc.AHRS;
-import edu.wpi.first.wpilibj.*;
+import edu.wpi.first.wpilibj.DriverStation;
+import edu.wpi.first.wpilibj.PowerDistribution;
+import edu.wpi.first.wpilibj.SPI;
+import edu.wpi.first.wpilibj.TimedRobot;
+import edu.wpi.first.wpilibj.PneumaticsControlModule;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.CommandScheduler;
 /**
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -20,7 +20,7 @@ public class Robot extends TimedRobot {
     public PowerDistribution pdp = new PowerDistribution();
     private RobotContainer m_robotContainer;
     private Command m_autonomousCommand;
-    public static final PneumaticsBase pneumaticsBase = PneumaticsBase.getForType(0, PneumaticsModuleType.CTREPCM);
+    public static final PneumaticsControlModule pneumaticsBase = new PneumaticsControlModule(0);
 
     /**
      * This function is run when the robot is first started up and should be used for any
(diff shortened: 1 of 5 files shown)


barel13 (approved):
Now we only need to fix this issue #37

src/main/java/frc/robot/Robot.java
```diff
@@ -5,10 +5,7 @@
 package frc.robot;
 
 import com.kauailabs.navx.frc.AHRS;
-import edu.wpi.first.wpilibj.DriverStation;
-import edu.wpi.first.wpilibj.PowerDistribution;
-import edu.wpi.first.wpilibj.SPI;
-import edu.wpi.first.wpilibj.TimedRobot;
+import edu.wpi.first.wpilibj.*;
```
katzuv: Don't use star imports, they clutter your namespace. Import only what you need.

src/main/java/frc/robot/Robot.java
```diff
@@ -23,6 +20,7 @@
     public PowerDistribution pdp = new PowerDistribution();
     private RobotContainer m_robotContainer;
     private Command m_autonomousCommand;
+    public static final PneumaticsBase pneumaticsBase = PneumaticsBase.getForType(0, PneumaticsModuleType.CTREPCM);
```
katzuv: You should use the [`PneumaticsControlModule`](https://first.wpi.edu/wpilib/allwpilib/docs/release/java/edu/wpi/first/wpilibj/PneumaticsControlModule.html) class directly.

src/main/java/frc/robot/subsystems/hood/Hood.java
```diff
@@ -3,14 +3,21 @@
...
 public class Hood extends SubsystemBase {
     private static Hood INSTANCE;
-    private final Solenoid angleChanger = new Solenoid(PneumaticsModuleType.CTREPCM, SOLENOID);
+    private final Solenoid angleChanger;
 
     private Hood() {
+        if (Robot.pneumaticsBase.checkSolenoidChannel(SOLENOID)) {
+            angleChanger = new Solenoid(PneumaticsModuleType.CTREPCM, SOLENOID);
+        } else {
+            angleChanger = null;
```
katzuv, line 20: Wouldn't that crash the code if i.e. `close()` is called?

katzuv (changes requested):
This is a great feature but I'm afraid it doesn't do what you want. The docs for `checkSolenoidChannel()` say it returns whether "a solenoid channel is valid", not if it's connected or not. "Valid" is probably in the valid range, as you can see in the implementation ([source](https://github.com/wpilibsuite/allwpilib/blob/10f19e6fc3ba982b08bb31da8c21c7b8241d0502/hal/src/main/native/sim/CTREPCM.cpp#L76-L78)):
```cpp
HAL_Bool HAL_CheckCTREPCMSolenoidChannel(int32_t channel) {
  return channel < kNumCTRESolenoidChannels && channel >= 0;
}
```

src/main/java/frc/robot/subsystems/hood/Hood.java
adamv17: We wouldn't use the close it just simplifies the testing process if something is not connected yet

katzuv: Will the testing be added in another branch?

src/main/java/frc/robot/Robot.java
```diff
@@ -23,6 +24,7 @@
     public PowerDistribution pdp = new PowerDistribution();
     private RobotContainer m_robotContainer;
     private Command m_autonomousCommand;
+    public static final PneumaticsControlModule pneumaticsBase = new PneumaticsControlModule(0);
```
katzuv, line 27: 1. The name now doesn't match, use either `PneumaticsControlModule` or `pcm`.
2. The default ID is `0` so you can use the simpler constructor: 
```suggestion
    public static final PneumaticsControlModule pneumaticsBase = new PneumaticsControlModule();
```

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
```diff
@@ -39,6 +42,11 @@ private Conveyor() {
         colorMatch.addColorMatch(NONE);
         cargoPositions.add(DriverStation.Alliance.Invalid.name());
         cargoPositions.add(DriverStation.Alliance.Invalid.name());
+        if (Utils.isBitSet(Robot.pneumaticsBase.getSolenoids(), SOLENOID)) {
```
katzuv, line 45: `getSolenoids()` is used in `Solenoid.get()` ([source](https://github.com/wpilibsuite/allwpilib/blob/5635f33a328144542e4ca7947c8c54614ce30de3/wpilibc/src/main/native/cpp/Solenoid.cpp#L50-L53)):
```cpp
bool Solenoid::Get() const {
  int currentAll = m_module->GetSolenoids();
  return (currentAll & m_mask) != 0;
}
```
So I think that unfortunately, that doesn't help either (or if it does, you can simply call `Solenoid`'s `get()`). What does `Solenoid.get()` return for a disconnected solenoid?

Also, just a quick question: Aren't double solenoids used in the robot?

BTW, I thought of having a "general test command" which you can run in the roboRIO's Test mode. This command will be a command group with basic commands for each subsystem (even use the BIT's one). This way, you can see whether mechanisms behave in the way you expect them to.
You can also do some more complicated tests. For example, the command for the shooter will print an error if the shooter didn't reach its setpoint 5 seconds after it started moving.
Moreover, think about getting user input and having that test command handle a "TODO list" for you. That way you don't need to remember what checks need to be done and which of these have finished.

katzuv (changes requested):

src/main/java/frc/robot/subsystems/conveyor/Conveyor.java
adamv17: 1. Honestly I am not sure anymore
2. We use single solenoids
3. Yes we will add bits for each subsystem, how would you handle a TODO list?

PR #71, opened 2022-02-07
Vision LEDs

adamv17:


barel13 (approved):

katzuv (approved):
Great! One suggestion I have is to rename the `LEDS` constant to something more specific because you might have other LEDs (🤩) on the robot in the future.

adamv17:
> Great! One suggestion I have is to rename the `LEDS` constant to something more specific because you might have other LEDs (🤩) on the robot in the future.

I think it's fine because it is under the Vision class

PR #87, opened 2022-02-09
Remove unused FireLog files

barel13:
As you can see from the [WebServer](https://github.com/Galaxia5987/robot-2022/blob/main/src/main/java/webapp/Webserver.java) we only use the [firescope.html](https://github.com/Galaxia5987/robot-2022/blob/main/src/main/deploy/firelog/firescope.html) with requires smoothie.js, vue.js and scope.js


katzuv (approved):
Great! Edit the PR's name to explain what's deleted, too.

adamv17 (approved):

PR #89, opened 2022-02-11
Add to swerve code

Se-gol:


katzuv:
Please edit the PR's title to be more descriptive. Also, rename the branch to adhere to the convention: `feature/branch-name`.

src/main/java/frc/robot/subsystems/drivetrain/SwerveDrive.java
```diff
@@ -11,6 +11,7 @@
 import edu.wpi.first.wpilibj2.command.SubsystemBase;
 import frc.robot.Constants;
 import frc.robot.Robot;
+import frc.robot.RobotContainer;
```
barel13: Remove unused import

src/main/java/frc/robot/subsystems/drivetrain/SwerveDrive.java
```diff
@@ -112,6 +113,71 @@ public void holonomicDrive(double forward, double strafe, double rotation) {
...
+     */
+    public void defaultHolonomicDrive(double forward, double strafe, double rotation) {
+        ChassisSpeeds speeds = fieldOriented ?
+                ChassisSpeeds.fromFieldRelativeSpeeds(forward, strafe, rotation, Robot.getAngle()) :
+                new ChassisSpeeds(forward, strafe, rotation);
+        setStates(kinematics.toSwerveModuleStates(speeds));
+    }
+
+    /**
+     * Move the swerve in the specified direction, rotation and velocity, but the velocities are relative to the error of the modules.
```
barel13: You don't need 3 functions for holonomicDrive, just create one function that chooses the velocity based on the rotation, vx and vy

src/main/java/frc/robot/subsystems/drivetrain/SwerveDrive.java
```diff
@@ -112,6 +113,71 @@ public void holonomicDrive(double forward, double strafe, double rotation) {
...
+     */
+    public boolean haveModulesReachedAngles(double forward, double strafe, double rotation) {
+        ChassisSpeeds chassisSpeeds = ChassisSpeeds.fromFieldRelativeSpeeds(
+                forward, strafe, rotation, Robot.getAngle());
+        SwerveModuleState[] states = kinematics.toSwerveModuleStates(chassisSpeeds);
+        for (int i = 0; i < 4; i++)
+            states[i] = SwerveModuleState.optimize(states[i], getModule(i).getAngle());
+        for (int i = 0; i < 4; i++) {
+            if (!(Math.abs(states[i].angle.minus(getModule(i).getAngle()).getDegrees()) < 10)) {
+                return false;
```
barel13:
```suggestion
        for (int i = 0; i < 4; i++) {
             states[i] = SwerveModuleState.optimize(states[i], getModule(i).getAngle());
            if (Math.abs(states[i].angle.minus(module[i].getAngle()).getDegrees()) > 10) {  // TODO: Remove the magic number
                return false;
```

barel13: You can also use for-each loop

src/main/java/frc/robot/subsystems/drivetrain/SwerveModule.java
```diff
@@ -94,8 +94,8 @@ public SwerveModule(SwerveModuleConfigBase config) {
         driveMotor.selectProfileSlot(1, 0);
         driveMotor.configIntegratedSensorInitializationStrategy(SensorInitializationStrategy.BootToZero);
 
-        driveMotor.configOpenloopRamp(Constants.SwerveModule.RAMP_RATE, Constants.TALON_TIMEOUT);
-        driveMotor.configClosedloopRamp(Constants.SwerveModule.RAMP_RATE, Constants.TALON_TIMEOUT);
+        driveMotor.configOpenloopRamp(0, Constants.TALON_TIMEOUT);
+        driveMotor.configClosedloopRamp(0, Constants.TALON_TIMEOUT);
```
barel13: Don't do that, just create 2 functions that let you choose the mode (motor-intenal or code defined)

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+
+import static frc.robot.Constants.SwerveDrive.VELOCITY_MULTIPLIER;
+
+public class OverpoweredDrive extends HolonomicDrive {
+    PIDController pidController = new PIDController(0.03, 0, 0) {{
+        enableContinuousInput(-180, 180);
+        setTolerance(5);
+    }};
+    private boolean newSetpoint = false;
+    private Rotation2d setpoint;
```
barel13: Why do you need that if you already have `headingController` in `SwerveDrive`?

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+    private boolean wait = false;
+    private double current = 0;
```
barel13, line 27: current what?

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+import static frc.robot.Constants.SwerveDrive.VELOCITY_MULTIPLIER;
+
+public class OverpoweredDrive extends HolonomicDrive {
+    PIDController pidController = new PIDController(0.03, 0, 0) {{
+        enableContinuousInput(-180, 180);
+        setTolerance(5);
+    }};
+    private boolean newSetpoint = false;
+    private Rotation2d setpoint;
+    private boolean wait = false;
```
barel13, line 26: make it internal in SwerveDrive

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+        double forward = speeds.vxMetersPerSecond;
+        double strafe = speeds.vyMetersPerSecond;
+        double magnitude = Math.hypot(forward, strafe);
+        double alpha = Math.atan2(strafe, forward);
+        if (magnitude == 0) current = 0;
+        current += magnitude / 10;
+        if (current > magnitude) current = magnitude;
+        forward = Math.cos(alpha) * current;
+        strafe = Math.sin(alpha) * current;
+        double rotation = speeds.omegaRadiansPerSecond;
```
barel13: I would extract it to another method (or even extend the `calculateVelocities()` method)

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+            // if there is no rotation after there was rotation, start the timer.
+            if (newSetpoint) {
+                newSetpoint = false;
+                timer.start();
+                timer.reset();
+            }
+            // if the time hasn't passed since there was no rotation after there was rotation, keep resetting the setpoint.
+            if (!timer.hasElapsed(0.2)) {
+                setpoint = Robot.getAngle();
+            }
```
barel13: Why do you need that?

src/main/java/frc/robot/subsystems/drivetrain/SwerveDrive.java
```diff
@@ -112,6 +113,71 @@ public void holonomicDrive(double forward, double strafe, double rotation) {
...
+    }
+
+    /**
+     * Check whether all modules have reached their desired angles.
+     *
+     * @param forward  the velocity on the X-axis. [m/s]
+     * @param strafe   the velocity on the Y-axis. [m/s]
+     * @param rotation the rotational velocity counter-clockwise positive. [rad/s]
+     */
+    public boolean haveModulesReachedAngles(double forward, double strafe, double rotation) {
```
barel13, line 164: If you use this function internally, you can just pass the states instead of calculating the velocities twice

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+
+
+        } else {
+            // if swerveDrive angles were reached don't wait
+            if (swerveDrive.haveModulesReachedAngles(forward, strafe, rotation))
+                wait = false;
+
+            // if you want acceleration from zero speed, and angles weren't reached
+            if (wait) {
```
barel13, line 76:
```suggestion
            else if (wait) {
```

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+                if (rotation == 0) {
+                    // if there is no rotation after there was rotation, start the timer.
+                    if (newSetpoint) {
+                        newSetpoint = false;
+                        timer.start();
+                        timer.reset();
+                    }
+
+                    // if the time hasn't passed since there was no rotation after there was rotation, keep resetting the setpoint.
+                    if (!timer.hasElapsed(0.2)) {
```
barel13:
```suggestion
                   else if (!timer.hasElapsed(0.2)) {
```

barel13: magic number

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+                        setpoint = Robot.getAngle();
+                    } else {
+                        swerveDrive.defaultHolonomicDrive(forward, strafe, pidController.calculate(Robot.getAngle().getDegrees(), setpoint.getDegrees()));
+                    }
+
+                } else {
+                    swerveDrive.defaultHolonomicDrive(forward, strafe, rotation * (1 + magnitude / (VELOCITY_MULTIPLIER * 2)));
```
barel13: what the hack is that? `rotation * (1 + magnitude / (VELOCITY_MULTIPLIER * 2))`

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+    private boolean wait = false;
+    private double current = 0;
+    private final Timer timer = new Timer();
+
+    public OverpoweredDrive(SwerveDrive swerveDrive, DoubleSupplier forwardSupplier, DoubleSupplier strafeSupplier, DoubleSupplier rotationSupplier) {
+        super(swerveDrive, forwardSupplier, strafeSupplier, rotationSupplier);
+    }
+
+    @Override
+    public void execute() {
```
barel13, line 35: You can simplify the logic here a lot more, also, you have a lot of unnecessary redundancy - please extract it to functions

barel13 (changes requested):

src/main/java/frc/robot/subsystems/drivetrain/SwerveDrive.java
```diff
@@ -94,9 +95,9 @@ public SwerveDriveKinematics getKinematics() {
     /**
      * Move the swerve in the specified direction, rotation and velocity.
      *
-     * @param forward  the velocity on the Y-axis. [m/s]
-     * @param strafe   the velocity on the X-axis. [m/s]
-     * @param rotation rhe rotational velocity. [rad/s]
+     * @param forward  the velocity on the X-axis. [m/s]
+     * @param strafe   the velocity on the Y-axis. [m/s]
```
barel13, line 98: is it always the x axis? what if we are in robot oriented mode?

barel13 (changes requested):

src/main/java/frc/robot/subsystems/drivetrain/commands/HolonomicDrive.java
```diff
@@ -40,15 +40,15 @@ public void execute() {
...
 
         // recalculate - update based on the angle and the magnitude
-        double alpha = Math.atan2(forward, strafe); // direction of movement
-        double magnitude = Math.hypot(forward, strafe) * Constants.SwerveDrive.VELOCITY_MULTIPLIER;
-        forward = Math.sin(alpha) * magnitude;
-        strafe = Math.cos(alpha) * magnitude;
+        double alpha = Math.atan2(strafe, forward); // direction of movement
+        double magnitude = Utils.deadband(Math.hypot(forward, strafe) * Constants.SwerveDrive.VELOCITY_MULTIPLIER, Constants.SwerveDrive.JOYSTICK_THRESHOLD);
+        forward = Math.cos(alpha) * magnitude;
+        strafe = Math.sin(alpha) * magnitude;
```
barel13, line 51: I don't think you need to change this, just replace the suppliers in the parameters. (as it would be better for robot-oriented and field-oriented

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
Se-gol: But then, image if haveModulesReachedAngles is true, then if won't go to the the else after the else if(wait), which does the actual driving.

barel13: So please separate the logics from the `holonimicDrive`'s extensions.

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
Se-gol: I can't do the else here. Imagine the robot is just rotating, then you stop holding the joystick and it's drifting a little. newSetpoint would be true. and then you want to record the current rotation for another 0.2 seconds. So the else wouldn't affect much, because immediately after, newSetpoint would be false and you would check the time. But, it makes less sense.

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
Se-gol: I'm adding to the magnitude to the rotation. The 2 divides it in half so I don't overcompensate.

src/main/java/frc/robot/subsystems/drivetrain/SwerveDrive.java
katzuv: What does "the velocities are relative to the error of the modules" mean?

src/main/java/frc/robot/subsystems/drivetrain/SwerveDrive.java
```diff
@@ -112,6 +113,71 @@ public void holonomicDrive(double forward, double strafe, double rotation) {
...
+        ChassisSpeeds chassisSpeeds = ChassisSpeeds.fromFieldRelativeSpeeds(
+                forward, strafe, rotation, Robot.getAngle());
+        SwerveModuleState[] states = kinematics.toSwerveModuleStates(chassisSpeeds);
+        for (int i = 0; i < 4; i++)
+            states[i] = SwerveModuleState.optimize(states[i], getModule(i).getAngle());
+        for (int i = 0; i < 4; i++) {
+            if (!(Math.abs(states[i].angle.minus(getModule(i).getAngle()).getDegrees()) < 10)) {
+                return false;
+            }
+        }
```
katzuv: Is it possible to put the two lines in the same loop?
Also, add curly brackets around line 171.

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+import frc.robot.Robot;
+import frc.robot.subsystems.drivetrain.SwerveDrive;
+
+import java.util.function.DoubleSupplier;
+
+import static frc.robot.Constants.SwerveDrive.VELOCITY_MULTIPLIER;
+
+public class OverpoweredDrive extends HolonomicDrive {
+    PIDController pidController = new PIDController(0.03, 0, 0) {{
+        enableContinuousInput(-180, 180);
```
katzuv: Don't you work with radians?

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+import edu.wpi.first.wpilibj.Timer;
+import frc.robot.Robot;
+import frc.robot.subsystems.drivetrain.SwerveDrive;
+
+import java.util.function.DoubleSupplier;
+
+import static frc.robot.Constants.SwerveDrive.VELOCITY_MULTIPLIER;
+
+public class OverpoweredDrive extends HolonomicDrive {
+    PIDController pidController = new PIDController(0.03, 0, 0) {{
```
katzuv: Magic numbers.

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+import edu.wpi.first.math.kinematics.ChassisSpeeds;
+import edu.wpi.first.wpilibj.Timer;
+import frc.robot.Robot;
+import frc.robot.subsystems.drivetrain.SwerveDrive;
+
+import java.util.function.DoubleSupplier;
+
+import static frc.robot.Constants.SwerveDrive.VELOCITY_MULTIPLIER;
+
+public class OverpoweredDrive extends HolonomicDrive {
```
katzuv, line 19: What does "overpowered" mean? Rename the command name to be more descriptive, and add documentation.

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,105 @@
...
+    PIDController pidController = new PIDController(0.03, 0, 0) {{
+        enableContinuousInput(-180, 180);
+        setTolerance(5);
```
katzuv: Magic number.

katzuv (changes requested):
Please document what the `OverpoweredDrive` command does. Thanks!

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.util.function.DoubleSupplier;
+
+import static frc.robot.Constants.SwerveDrive.VELOCITY_MULTIPLIER;
+
+/*
+    OverpoweredDrive is an assisted driving command that provides a smoother driving experience.
+    It corrects heading drifting, slows acceleration, and waits for wheels to reach their desired rotations before speeding.
+ */
+public class OverpoweredDrive extends HolonomicDrive {
```
barel13, line 19: This command is a copy of `DriveSlowAcceleration`, the only difference is that you use the the heading controller in the command, and in `DriveSlowAcceleration` we use the heading controller in the SwerveDrive

barel13 (changes requested):

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
```diff
@@ -0,0 +1,111 @@
...
+    private final PIDController pidController = new PIDController(Constants.SwerveDrive.HEADING_KP, 0, 0) {{
+        enableContinuousInput(-Math.PI, Math.PI);
+        setTolerance(Math.toRadians(Constants.SwerveDrive.ALLOWABLE_HEADING_ERROR));
+    }};
+    private boolean newSetpoint = false;
+    private Rotation2d setpoint;
+    private boolean wait = false;
+    private double current = 0;
```
barel13, line 27: Why don't you use SlewRateLimiter?

src/main/java/frc/robot/subsystems/drivetrain/SwerveModule.java
```diff
@@ -253,6 +250,17 @@ private void configPID(double kp, double ki, double kd, double kf) {
...
 
+
+    private void selectTuneDownMode(boolean isUserDefined) {
+        if (isUserDefined) {
+            driveMotor.configOpenloopRamp(0, Constants.TALON_TIMEOUT);
+            driveMotor.configClosedloopRamp(0, Constants.TALON_TIMEOUT);
+        } else {
+            driveMotor.configOpenloopRamp(Constants.SwerveModule.RAMP_RATE, Constants.TALON_TIMEOUT);
+            driveMotor.configClosedloopRamp(Constants.SwerveModule.RAMP_RATE, Constants.TALON_TIMEOUT);
+        }
```
katzuv:
```suggestion
        double ramp;
        if (isUserDefined) {
            ramp = 0;
            } else {
            ramp = Constants.SwerveModule.RAMP_RATE;
        }
        driveMotor.configOpenloopRamp(ramp, Constants.TALON_TIMEOUT);
        driveMotor.configClosedloopRamp(ramp, Constants.TALON_TIMEOUT);
```
Or:
```suggestion
        double ramp = isUserDefined ? 0 : Constants.SwerveModule.RAMP_RATE;
        driveMotor.configOpenloopRamp(ramp, Constants.TALON_TIMEOUT);
        driveMotor.configClosedloopRamp(ramp, Constants.TALON_TIMEOUT);
```
Also, if you don't read the error code returned from the config, there is no need to set the timout.

katzuv (changes requested):

katzuv:
Also, please see my comment [here](https://github.com/Galaxia5987/robot-2022/pull/89#issuecomment-1036313565) and consider renaming `OverpoweredDrive` to `AssistedDrive`.

src/main/java/frc/robot/subsystems/drivetrain/commands/OverpoweredDrive.java
Se-gol: It is possible and it's not too different. But, I like this better for the fact that it doesn't go above the magnitude. Which slew can when going from high to low. Which makes it feel less responsive.

PR #97, opened 2022-02-21
Trying to fix shoot cargo logic

vichik123:
There appears to be a problem with the `ShootCargo` command, so we are trying to fix it. Unfortunately I can't seem to find the problem, although the first commit made to this pull request solved another.


katzuv:
> although the first commit made to this pull request solved another.

If so, I recommend renaming this branch and PR to show that it solves this specific bug. Then, merge this PR and open another that will solve your currently-unsolved problem.
This will make sure the bug is fixed faster and the diff will be smaller. Generally, branches should fix or add only one thing.

Note that if this branch fixes a bug it should be prefixed with `hotfix/` and not with `feature/`. Also, "fix logic" is pretty vague, try to find a more specific name for the branch. 

Good luck! 😸

PR #99, opened 2022-03-13
Feature/add safeties

Se-gol:
Add vision safeties for competition.

Every Command gets a special case for when the vision can't see the target.

Adjust to target commands are switched to angle by odometry.

Shooting commands are switched to distance from odometry.

Also in auto.

Constant for whether or not we're playing without vision.

### DIFF
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -136,7 +136,7 @@ public final class Constants {
             put(5.65, 4550.0);
             put(99999.0, 4550.0);
         }};
-        public static final double TARMAC_VELOCITY = 3700;
+        public static final double TARMAC_VELOCITY = 3700; // [rpm]
 
         public static TalonFXConfiguration getConfiguration() {
             final TalonFXConfiguration configuration = new TalonFXConfiguration();
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -97,6 +97,7 @@ public final class Constants {
         public static final double NEUTRAL_DEADBAND = 0.1; // [%]
         public static final double OUTPUT_MULTIPLIER = 0.1; // Multiplies the output for manual control in the bits. [%]
         public static final double OUTTAKE_POWER = 0.2; // Power to give to the shooter when taking balls out. [%]
+        public static final double LOW_GOAL_VELOCITY = 2500;
         public static final double RECOMMENDED_ACCELERATION_TIME = 1.3; // Recommended time for the shooter to get to it's setpoint. [s]
         public static final double CARGO_OFFSET = 0; // Desired offset from the middle of the target where you want the cargo to hit. [m]
 
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -282,7 +282,7 @@ public final class Constants {
         public static final double POWER_TO_VELOCITY_RATIO = -3 / 16.0; // Ratio of power to velocity. [% / m/s]
         public static final double TIME_BETWEEN_RUNS = 1.7; // time intake will wait before toggling the retractor (for testing only). [s]
 
-        public static final WebConstant DEFAULT_POWER = WebConstant.of("Intake", "power", 1); // power intake will receive on the basic command. [%]
+        public static final WebConstant DEFAULT_POWER = WebConstant.of("Intake", "power", 0.8); // power intake will receive on the basic command. [%]
     }
 
     public static class Hood {
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -34,8 +34,8 @@ import java.util.function.DoubleSupplier;
 import java.util.function.Supplier;
 
 public class RobotContainer {
-    public static boolean playWithoutVision = false;
-    public static boolean hardCodedVelocity = false;
+    public static final boolean playWithoutVision = false;
+    public static final boolean hardCodedVelocity = false;
 
     // The robot's subsystems and commands are defined here...
     public static LedSubsystem ledSubsystem = new LedSubsystem();
@@ -99,7 +99,7 @@ public class RobotContainer {
             }, shooter).withInterrupt(Xbox.rt::get));
             Xbox.a.whileHeld(new BackAndShootCargoSort(shooter, hood, conveyor, flap,
                     () -> Constants.Conveyor.SHOOT_POWER,
-                    distanceFromTarget, photonVisionModule::hasTargets, swerve::getOdomDistance));
+                    distanceFromTarget, photonVisionModule::hasTargets, swerve::getOdometryDistance));
 
             Xbox.leftPov.whileActiveOnce(new InstantCommand(hood::toggle));
             Xbox.rightPov.whileActiveOnce(new InstantCommand(helicopter::toggleStopper));
@@ -109,7 +109,7 @@ public class RobotContainer {
             Xbox.rt.whileActiveContinuous(new BackAndShootCargo(
                     shooter, hood, conveyor, flap,
                     () -> Constants.Conveyor.SHOOT_POWER,
-                    distanceFromTarget, photonVisionModule::hasTargets, swerve::getOdomDistance));
+                    distanceFromTarget, photonVisionModule::hasTargets, swerve::getOdometryDistance));
             Xbox.lt.whileActiveContinuous(new PickUpCargo(conveyor, flap, intake, Constants.Conveyor.DEFAULT_POWER.get(), Constants.Intake.DEFAULT_POWER::get));
             Xbox.lb.whileHeld(new Outtake(intake, conveyor, flap, shooter, hood, () -> false));
             Xbox.rb.whileHeld(new Convey(conveyor, -Constants.Conveyor.SHOOT_POWER));
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -8,10 +8,7 @@ import edu.wpi.first.wpilibj2.command.*;
 import edu.wpi.first.wpilibj2.command.button.JoystickButton;
 import edu.wpi.first.wpilibj2.command.button.Trigger;
 import frc.robot.auto.TaxiFromLowRightPickShootPickShoot;
-import frc.robot.commandgroups.BackAndShootCargoSort;
-import frc.robot.commandgroups.OneBallOuttake;
-import frc.robot.commandgroups.Outtake;
-import frc.robot.commandgroups.PickUpCargo;
+import frc.robot.commandgroups.*;
 import frc.robot.commandgroups.bits.RunAllBits;
 import frc.robot.subsystems.conveyor.Conveyor;
 import frc.robot.subsystems.conveyor.commands.Convey;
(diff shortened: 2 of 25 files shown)


src/main/java/frc/robot/subsystems/drivetrain/commands/DriveAndAdjustWithVisionExperimental.java
```diff
@@ -0,0 +1,156 @@
...
+    private final DoubleSupplier distanceSupplier;
+    private final BooleanSupplier hasTarget;
+    private boolean newSetpoint = false;
+    private Rotation2d setpoint;
+    private boolean wait = false;
+    private double current = 0;
+    private Rotation2d target;
+    private boolean last = false;
+
+    public DriveAndAdjustWithVisionExperimental(SwerveDrive swerveDrive, DoubleSupplier forwardSupplier, DoubleSupplier strafeSupplier, DoubleSupplier rotationSupplier, DoubleSupplier yawSupplier, BooleanSupplier condition, DoubleSupplier distanceSupplier, BooleanSupplier hasTarget) {
```
adamv17, line 39: Can you explain this command?

src/test/java/Test.java
```diff
@@ -0,0 +1,18 @@
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.math.geometry.Translation2d;
+
+public class Test {
+    @org.junit.Test
+    public void test() {
```
adamv17: Either delete this or better yet create a function with this logic so that we can check this in the future

src/main/java/frc/robot/Constants.java
```diff
@@ -136,6 +136,7 @@
             put(5.65, 4550.0);
             put(99999.0, 4550.0);
         }};
+        public static final double TARMAC_VELOCITY = 3700;
```
barel13: Units

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -35,6 +34,8 @@
 import java.util.function.Supplier;
 
 public class RobotContainer {
+    public static boolean playWithoutVision = false;
+    public static boolean hardCodedVelocity = false;
```
barel13:
```suggestion
    public static final boolean playWithoutVision = false;
    public static final boolean hardCodedVelocity = false;
```

src/main/java/frc/robot/commandgroups/BackAndShootCargoSort.java
```diff
@@ -15,8 +16,8 @@ public BackAndShootCargoSort(Shooter shooter,
                                  Conveyor conveyor,
                                  Flap flap,
                                  DoubleSupplier conveyorPower,
-                                 DoubleSupplier distanceFromTarget) {
+                                 DoubleSupplier distanceFromTarget, BooleanSupplier hasTarget, DoubleSupplier odomDistance) {
```
barel13: You can remove `hasTarget` and `odomDistance` and just pass the value from the odometry if the robot doesn't see the target. (It's also applies to other commands, I just don't want to copy & paste)

src/main/java/frc/robot/subsystems/drivetrain/SwerveDrive.java
```diff
@@ -342,6 +342,10 @@ public void lock() {
         modules[1].setAngle(Rotation2d.fromDegrees(-45));
     }
 
+    public double getOdomDistance() {
```
barel13: documentation, and please rename from odom to odometry

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
```diff
@@ -0,0 +1,61 @@
...
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.wpilibj2.command.CommandBase;
+import frc.robot.Constants;
+import frc.robot.Robot;
+import frc.robot.subsystems.drivetrain.SwerveDrive;
+
+import java.util.function.BooleanSupplier;
+import java.util.function.DoubleSupplier;
+
+public class AdjustToTargetOnCommand extends CommandBase {
```
barel13, line 13: Can be renamed to `AdjustToTarget` (it's in commands package so its obvious that it's a command)

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
```diff
@@ -0,0 +1,61 @@
...
+    private final SwerveDrive swerveDrive;
+    private final DoubleSupplier yawSupplier;
+    private final PIDController adjustController = new PIDController(Constants.SwerveDrive.ADJUST_CONTROLLER_KP.get(), 0, 0) {{
+        enableContinuousInput(-Math.PI, Math.PI);
+        setTolerance(Constants.SwerveDrive.ADJUST_CONTROLLER_TOLERANCE);
+    }};
```
barel13, line 19: Please put the `{{ }}` in the constructor

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
```diff
@@ -0,0 +1,61 @@
...
+    private final BooleanSupplier hasTarget;
+    private Rotation2d target;
+
+
+    public AdjustToTargetOnCommand(SwerveDrive swerveDrive, DoubleSupplier yawSupplier, BooleanSupplier hasTarget) {
```
barel13, line 24: The yaw is the absolute angle or the angle relative to the robot

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
```diff
@@ -0,0 +1,61 @@
...
+        this.swerveDrive = swerveDrive;
+        this.yawSupplier = yawSupplier;
+        this.hasTarget = hasTarget;
+        target = Robot.getAngle();
```
barel13, line 28: Needs to be initialized in `initialize` (you don't have the gyro at compile time)

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
```diff
@@ -0,0 +1,61 @@
...
+            var robotAngle = Robot.getAngle();
+            var robotPose = swerveDrive.getPose().getTranslation();
+            var hubPose = Constants.Vision.HUB_POSE.getTranslation();
+            var poseRelativeToTarget = hubPose.minus(robotPose);
+            var value = robotAngle.plus(new Rotation2d(
+                    Math.atan2(
+                            poseRelativeToTarget.getY(),
+                            poseRelativeToTarget.getX()
+                    )));
+            target = value;
```
barel13:
```suggestion
            target = robotAngle.plus(new Rotation2d(
                    Math.atan2(
                            poseRelativeToTarget.getY(),
                            poseRelativeToTarget.getX()
                    )));
```

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class AdjustToTargetOnCommand extends CommandBase {
+    private final SwerveDrive swerveDrive;
+    private final DoubleSupplier yawSupplier;
+    private final PIDController adjustController = new PIDController(Constants.SwerveDrive.ADJUST_CONTROLLER_KP.get(), 0, 0) {{
+        enableContinuousInput(-Math.PI, Math.PI);
+        setTolerance(Constants.SwerveDrive.ADJUST_CONTROLLER_TOLERANCE);
+    }};
+    private final BooleanSupplier hasTarget;
+    private Rotation2d target;
```
barel13, line 21: `target` can be stored directly in `adjustController` as the setpoint

src/main/java/frc/robot/subsystems/drivetrain/commands/DriveAndAdjustWithVisionExperimental.java
```diff
@@ -0,0 +1,156 @@
...
+import frc.robot.Constants;
+import frc.robot.Robot;
+import frc.robot.subsystems.drivetrain.SwerveDrive;
+
+import java.util.function.BooleanSupplier;
+import java.util.function.DoubleSupplier;
+
+import static frc.robot.Constants.SwerveDrive.VELOCITY_MULTIPLIER;
+
+public class DriveAndAdjustWithVisionExperimental extends HolonomicDrive {
```
barel13, line 16: Why not just extend from OverpoweredDrive and have most of it implemented?

barel13 (changes requested):

src/main/java/frc/robot/subsystems/drivetrain/commands/DriveAndAdjustWithVisionExperimental.java
Se-gol: It's the same as the DriveAndAdjust, but it takes only one input from the vision, on init. Unlike DriveAndAdjust, which updates the target angle every some milliseconds .

src/test/java/Test.java
Se-gol: yes, I didn't mean to leave it in.

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
Se-gol: It's not AdjustToTargetCommand it's AdjustToTargetOnCommand.

src/main/java/frc/robot/commandgroups/BackAndShootCargoSort.java
Se-gol: I see what you're saying. To make it like an optional double, or something like that. But, I like it less. Also, it somewhat ruins the idea of the playWithoutVision boolean, which if true, will make it so hasTarget will always return false.

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
Se-gol: I do already initialize it in initialize. I could have just as well removed that line at 28. it's just for safety.

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
Se-gol: I don't see a reason to do that. Also, I haven't seen setSetpoint working yet.

src/main/java/frc/robot/subsystems/drivetrain/commands/DriveAndAdjustWithVisionExperimental.java
Se-gol: I don't see how I would go about on doing that, and I believe it would only complicate things.

src/main/java/frc/robot/subsystems/drivetrain/commands/AdjustToTargetOnCommand.java
Se-gol: The yaw is the relative yaw to the hub. The target I calculate is the absolute angle.

PR #103, opened 2022-03-26
Commit all, explain later

Se-gol:
It's not pretty, I'm sorry.

In teleop, I have a default command for the shooter that tries to reach the double, `setpointVelocity` in `RobotContainer`.

In `SwerveDrive` periodic, I am updating the `setpointVelocity` based on odometry and vision (when available).

When shooting, I lower the flap, wait for setpoint, activate conveyor, and when the second ball comes, I stop him for 0.3 seconds.

The boolean, `shooting`, in `RobotContainer` makes it so that when shooting, the setpoint stays the same.

Additionally, I have created a button for the driver to make the shooting as quick as it could be (by Paulo's request).

The button, first adjusts using odometry, than fixes itself with vision, and then shoots up to 2 balls. All while holding only one button.

In auto, I have a `reachVelocityByDistance` command that simply sets the shooter velocity.

When ready to shoot I have a `confirmShooting` function that activates the conveyor and stops for the second ball.


src/main/java/frc/robot/commandgroups/JustShoot.java
```diff
@@ -0,0 +1,26 @@
...
+import frc.robot.subsystems.conveyor.Conveyor;
+import frc.robot.subsystems.conveyor.commands.Convey;
+import frc.robot.subsystems.flap.Flap;
+import frc.robot.subsystems.hood.Hood;
+
+public class JustShoot extends SequentialCommandGroup {
+    public JustShoot(Conveyor conveyor, Flap flap, Hood hood) {
+        addCommands(new Convey(conveyor, -0.25).withTimeout(0.075));
+        addCommands(new InstantCommand(() -> RobotContainer.shooting = true));
+//        if (RobotContainer.hardCodedVelocity)
```
adamv17: Why are these comments necessary?

src/main/java/frc/robot/commandgroups/OdometryAdjust.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Override
+    public void execute() {
+        double rotation = adjustController.calculate(Robot.getAngle().getRadians(), target.getRadians());
+        swerveDrive.holonomicDrive(0, 0, rotation);
+    }
+
+    @Override
+    public boolean isFinished() {
+        return Math.abs(Robot.getAngle().minus(target).getDegrees()) < 2;
```
adamv17: Can you make this a constants plz, this is true for a few more classes

src/main/java/frc/robot/commandgroups/OneButtonAdjustAndShoot.java
```diff
@@ -0,0 +1,36 @@
...
+import frc.robot.subsystems.hood.Hood;
+
+public class OneButtonAdjustAndShoot extends SequentialCommandGroup {
+    public OneButtonAdjustAndShoot(SwerveDrive swerve, Conveyor conveyor, Flap flap, Hood hood) {
+        addCommands(new InstantCommand(flap::allowShooting));
+
+        addCommands(new ParallelCommandGroup(
+                new Convey(conveyor, -0.25).withTimeout(0.075),
+                //if (!RobotContainer.hasTarget.getAsBoolean())
+                new OdometryAdjust(swerve)
```
adamv17, line 23: What happens if the odometry is incorrect? Is there some kind of way to overrule this

src/main/java/frc/robot/commandgroups/OneButtonAdjustAndShoot.java
```diff
@@ -0,0 +1,36 @@
...
+        ).withTimeout(1));
+//        if (RobotContainer.hasTarget.getAsBoolean() && !RobotContainer.playWithoutVision)
+        addCommands(new VisionAdjust(swerve).withTimeout(0.5));
+        addCommands(new InstantCommand(() -> RobotContainer.shooting = true));
+//        if (RobotContainer.hardCodedVelocity)
+//            addCommands(new InstantCommand(() -> hood.setSolenoid(RobotContainer.hardCodedDistance < Constants.Hood.DISTANCE_FROM_TARGET_THRESHOLD ? Hood.Mode.ShortDistance : Hood.Mode.LongDistance)));
+//        else {
+//            if (RobotContainer.hasTarget.getAsBoolean() && !RobotContainer.playWithoutVision)
+        addCommands(new InstantCommand(() -> hood.setSolenoid(RobotContainer.Suppliers.distanceSupplier.getAsDouble() < Constants.Hood.DISTANCE_FROM_TARGET_THRESHOLD ? Hood.Mode.ShortDistance : Hood.Mode.LongDistance)));
```
adamv17, line 35: I suggest you make a hood command that is an instant command that does just this, as you use it in multiple places

src/main/java/frc/robot/subsystems/conveyor/commands/ConveyToShooter.java
```diff
@@ -30,98 +30,98 @@ public ConveyToShooter(Conveyor conveyor, BooleanSupplier preFlapSupplier, Doubl
...
-        timer.stop();
-        delayTimer.stop();
-        conveyor.setPower(0);
-    }
-
-    @Override
-    public boolean isFinished() {
-        return super.isFinished();
-    }
+//
```
adamv17: Just delete the commented out code, everything is on git

adamv17 (changes requested):
This seems like it might work although as of right now if one of the pieces doesn't work we either don't shoot or miss

src/main/java/frc/robot/commandgroups/JustShoot.java
Se-gol: These comments replace the hard coded and safeties from before. It might be good to uncomment, but first we need to see the regular version works

src/main/java/frc/robot/commandgroups/OneButtonAdjustAndShoot.java
Se-gol: I don't think the odometry could ever get so bad to the point it's unusable. I was thinking to use odometry only when you don't see the target but I'm thinking this will be more accurate. We could put a button that does that. In any way, after odometry comes the vision fix which should make it reliable either way.

src/main/java/frc/robot/commandgroups/JustShoot.java
adamv17: I think we should just uncomment it shouldn't change the logic of the code if these safeties are turned off

src/main/java/frc/robot/commandgroups/OneButtonAdjustAndShoot.java
adamv17: Okay we'll have to see it in drivers practice but it might be useful to split it a bit.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -119,39 +107,21 @@ private void configureDefaultCommands() {
...
             Xbox.y.whenPressed(new InstantCommand(() -> warmUpShooting = !warmUpShooting));
             Xbox.a.whileHeld(() -> playWithoutVision = true).whenReleased(() -> playWithoutVision = false);
 
             Xbox.leftPov.whileActiveOnce(new InstantCommand(hood::toggle));
             Xbox.downPov.whileActiveOnce(new LowGoalShot(shooter, flap, hood));
             Xbox.upPov.whileActiveContinuous(new JoystickPowerHelicopter(helicopter, () -> -0.1));
 
-            Xbox.rt.whileActiveContinuous(new BackAndShootCargo(
-                    shooter, hood, conveyor, flap,
-                    () -> 0)).whenInactive(() -> shooting = false);
```
vichik123, line 137: Do you think it's a good idea to have Ma'ayan shoot the cargo? If the adjustment doesn't work it could cause problems.

src/main/java/frc/robot/commandgroups/OneButtonAdjustAndShoot.java
```diff
@@ -0,0 +1,45 @@
...
+        addCommands(new InstantCommand(() -> RobotContainer.ledSubsystem.setCurrentLedMode(LedSubsystem.LedMode.VISION_ADJUST))); // LEDs
+
+//        if (RobotContainer.hasTarget.getAsBoolean() && !RobotContainer.playWithoutVision)
+        addCommands(new VisionAdjust(swerve).withTimeout(0.5));
+        addCommands(new InstantCommand(() -> RobotContainer.shooting = true));
+//        if (RobotContainer.hardCodedVelocity)
+//            addCommands(new InstantCommand(() -> hood.setSolenoid(RobotContainer.hardCodedDistance < Constants.Hood.DISTANCE_FROM_TARGET_THRESHOLD ? Hood.Mode.ShortDistance : Hood.Mode.LongDistance)));
+//        else {
+//            if (RobotContainer.hasTarget.getAsBoolean() && !RobotContainer.playWithoutVision)
+        addCommands(new InstantCommand(() -> hood.setSolenoid(RobotContainer.Suppliers.distanceSupplier.getAsDouble() < Constants.Hood.DISTANCE_FROM_TARGET_THRESHOLD ? Hood.Mode.ShortDistance : Hood.Mode.LongDistance)));
```
vichik123: I don't really like the use of the `addCommands` function so many times. Why not just use it once?

vichik123 (commented):
This looks good, much cleaner. However, I would recommend removing as much commented code as possible.

src/main/java/frc/robot/commandgroups/bits/TestBallFlow.java
```diff
@@ -24,14 +20,6 @@ public TestBallFlow(Hood hood, Flap flap, Conveyor conveyor, Intake intake, Shoo
...
         this.shooter = shooter;
         this.conveyor = conveyor;
-        ShootCargo shootCargo = new ShootCargo(shooter, hood, conveyor, flap, Constants.Conveyor.DEFAULT_POWER::get);
-        PickUpCargo pickUpCargo = new PickUpCargo(conveyor, flap, intake, Constants.Conveyor.DEFAULT_POWER.get(), () -> 0.5);
-
-        addCommands(
-                pickUpCargo.withInterrupt(pickUpCargoIsFinished),
-                new WaitCommand(5),
-                shootCargo.withInterrupt(shootCargoIsFinished)
-        );
```
barel13, line 34: Why did you remove the whole point of this command?

src/main/java/frc/robot/subsystems/conveyor/commands/ConveyCargo.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public void initialize() {
+        timer.reset();
+        timer.stop();
+    }
+
+    @Override
+    public void execute() {
```
barel13, line 27: I'm not sure it would work the way you wanted (and I certain that it can be implemented in an easier way)

src/main/java/frc/robot/subsystems/drivetrain/commands/DriveAndAdjustWithVision.java
```diff
@@ -60,8 +60,8 @@ public void execute() {
         if (magnitude == 0) current = 0;
         current += magnitude / 5;
         if (current > magnitude) current = magnitude;
-        forward = Math.cos(alpha) * current;
-        strafe = Math.sin(alpha) * current;
+        forward = Math.cos(alpha) * magnitude;
+        strafe = Math.sin(alpha) * magnitude;
```
barel13, line 64: So you don't use the `current` variable?

src/main/java/frc/robot/subsystems/hood/commands/HoodCommand.java
```diff
@@ -2,32 +2,30 @@
 
 import edu.wpi.first.wpilibj.Timer;
 import edu.wpi.first.wpilibj2.command.CommandBase;
-import frc.robot.Constants;
-import frc.robot.RobotContainer;
 import frc.robot.subsystems.hood.Hood;
 
 public class HoodCommand extends CommandBase {
```
barel13, line 7: You can remove the class, there is no need to just comment classes as it would make more troubles  for other commands who use this command...

src/main/java/frc/robot/subsystems/shooter/commands/BackAndShootCargo.java
```diff
@@ -1,11 +1,7 @@
 package frc.robot.subsystems.shooter.commands;
```
barel13, line 1: Again, you just deleted the whole command

src/main/java/frc/robot/subsystems/shooter/commands/Shoot.java
```diff
@@ -73,7 +69,7 @@ public static double getSetpointVelocity(double distance) {
         double t = (distance - prevMeasuredDistance) / (nextMeasuredDistance - prevMeasuredDistance);
         return (1 - t) * y1 + t * y2;
     }
-
+/*
```
barel13, line 72: Why?

src/main/java/frc/robot/subsystems/shooter/commands/bits/TryVelocities.java
```diff
@@ -36,9 +34,9 @@ public TryVelocities(Conveyor conveyor, Shooter shooter, Hood hood, Flap flap,
         this.hood = hood;
         this.flap = flap;
         this.distanceFromTarget = distanceFromTarget;
-        addCommandsForNextVelocity();
+//        addCommandsForNextVelocity();
```
barel13, line 37: ...

barel13 (changes requested):

src/main/java/frc/robot/commandgroups/bits/TestBallFlow.java
Se-gol: I commented what gave me errors, what's important can be brought back

src/main/java/frc/robot/RobotContainer.java
Se-gol: Well, I also thought about it. Paulo recommended it and I also think it helps with the flow of the game. All Maayan has to do is press and hold once until balls go in. We'll see how well that, and the adjustment work. If the adjustment doesn't work we'll find safeties or split it. And if Maayan prefers it, we'll let you do it

src/main/java/frc/robot/commandgroups/OneButtonAdjustAndShoot.java
Se-gol: It just helps me visually separate them from each other, but I guess addCommands to everything isn't too bad either

src/main/java/frc/robot/subsystems/drivetrain/commands/DriveAndAdjustWithVision.java
Se-gol: Usually, it's fine. But, because we don't fully stop until wheels are adjusted, and because the speed multiplier is so high, and because of the x0.5 speed on trigger. I noticed it could cause some problems.

src/main/java/frc/robot/subsystems/shooter/commands/Shoot.java
Se-gol: We don't need it

src/main/java/frc/robot/subsystems/conveyor/commands/ConveyCargo.java
Se-gol: What wouldn't work and what's not easy about it? The first ball I shoot straight away. As soon as I see a ball went in and out of the pre flap or into the post flap I know the second ball is next, and then I wait only for it at the pre flap for 0.3 seconds

katzuv:
> The boolean, `shooting`, in `RobotContainer` makes it so that _when shooting, the setpoint stays the same_.

Why is the emphasized text needed?

